PR #24, opened 2021-05-21
test: add a test to demonstrate an issue with soft deletes in GraphQL

simpers:
### Contributor checklist
No fix but a test that once fixed _should hopefully_ compile and test the functionality.

Might want to be able to check that is is indeed still left in the database too?


zachdaniel:
Beautiful

PR #32, opened 2022-05-05
improvement: setup generate_object? setting on resource

michaelst:
### Contributor checklist

### DIFF
--- a/lib/resource/resource.ex
+++ b/lib/resource/resource.ex
@@ -2122,8 +2122,6 @@ defmodule AshGraphql.Resource do
         name: Macro.camelize(to_string(type)),
         __reference__: ref(__ENV__)
       }
-    else
-      []
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/resource/resource.ex
```diff
@@ -2093,24 +2103,28 @@ defmodule AshGraphql.Resource do
...
+        description: Ash.Resource.Info.description(resource),
+        interfaces: interfaces,
+        fields: fields(resource, api, schema),
+        identifier: type,
+        module: schema,
+        name: Macro.camelize(to_string(type)),
+        __reference__: ref(__ENV__)
+      }
+    else
+      []
```
zachdaniel, line 2126: can we return `nil` here instead? Otherwise, looks great!

zachdaniel (changes requested):

lib/resource/resource.ex
michaelst: done

zachdaniel:

PR #33, opened 2022-05-10
fix: Handle error if multitenant resource was fetched without tenant being set

dolfinus:
I've played a bit with [example project|https://github.com/ash-project/ash_example], and found an issue.

I've created a simple ticket with id `99bb4bec-b8ea-494e-83da-1f706c44a1e1`, and then tried to execute this query:
```graphql
query {
  getTicket(id: "99bb4bec-b8ea-494e-83da-1f706c44a1e1") {
    id
    description
    response
    representative {
      id
    }
    reporter {
      id
    }
    subject
    status
  }
}
```

It fails because tenant is not set in the context:
[old.log](https://github.com/ash-project/ash_graphql/files/8665107/old.log)

But instead of returning `Something went wrong` in `errors` field, the request is failed with 500 error.

I've changed `dataloader` behavior from raising an exception to returning a tuple `{:error, error}`, so this error could be handled by existing `to_resolution` function.

Now fields which cannot be fetched are not causing all the query to fail, `nil` is returned instead:
```json
{
  "data": {
    "getTicket": {
      "description": null,
      "id": "99bb4bec-b8ea-494e-83da-1f706c44a1e1",
      "reporter": null,
      "representative": null,
      "response": null,
      "status": "NEW",
      "subject": "aasbc"
    }
  },
  "errors": [
    {
      "locations": [
        {
          "column": 5,
          "line": 9
        }
      ],
      "message": "Something went wrong. Unique error id: `e47079eb-223b-432b-8fb4-c01714dc9925`",
      "path": [
        "getTicket",
        "reporter"
      ]
    },
    {
      "locations": [
        {
          "column": 5,
          "line": 6
        }
      ],
      "message": "Something went wrong. Unique error id: `f6a6a465-4164-4f67-9782-17aaca439cce`",
      "path": [
        "getTicket",
        "representative"
      ]
    }
  ]
}
```

But original error message is still printing to the log:
[now.log](https://github.com/ash-project/ash_graphql/files/8665105/now.log)

Also it is now possible to pass custom dataloader options as third argument of `AshGraphql.add_context`.

### Contributor checklist

### DIFF
--- a/lib/ash_graphql.ex
+++ b/lib/ash_graphql.ex
@@ -264,7 +264,8 @@ defmodule AshGraphql do
   end
 
   def add_context(ctx, apis, options \\ []) do
-    empty_dataloader = Dataloader.new(options ++ [get_policy: :tuples])
+    options = Keyword.put(options, :get_policy, :tuples)
+    empty_dataloader = Dataloader.new(options)
 
     dataloader =
       apis
--- a/lib/ash_graphql.ex
+++ b/lib/ash_graphql.ex
@@ -263,11 +263,13 @@ defmodule AshGraphql do
     end)
   end
 
-  def add_context(ctx, apis) do
+  def add_context(ctx, apis, options \\ []) do
+    empty_dataloader = Dataloader.new(options ++ [get_policy: :tuples])
+
     dataloader =
       apis
       |> List.wrap()
-      |> Enum.reduce(Dataloader.new(), fn {api, _registry}, dataloader ->
+      |> Enum.reduce(empty_dataloader, fn {api, _registry}, dataloader ->
         Dataloader.add_source(
           dataloader,
           api,
--- a/lib/graphql/resolver.ex
+++ b/lib/graphql/resolver.ex
@@ -932,6 +932,10 @@ defmodule AshGraphql.Graphql.Resolver do
     child_complexity + 1
   end
 
+  def fetch_dataloader(loader, api, batch_key, parent) do
+    to_resolution(Dataloader.get(loader, api, batch_key, parent))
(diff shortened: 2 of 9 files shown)


zachdaniel:
Looks like a great addition! I'll review and merge this evening or tomorrow. Thanks :D

lib/ash_graphql.ex
```diff
@@ -263,11 +263,13 @@ defmodule AshGraphql do
     end)
   end
 
-  def add_context(ctx, apis) do
+  def add_context(ctx, apis, options \\ []) do
+    empty_dataloader = Dataloader.new(options ++ [get_policy: :tuples])
```
zachdaniel, line 267: I think we should do `Keyword.put(options, :get_policy, :tuples)` here.

lib/graphql/resolver.ex
```diff
@@ -932,6 +932,10 @@ defmodule AshGraphql.Graphql.Resolver do
     child_complexity + 1
   end
 
+  def fetch_dataloader(loader, api, batch_key, parent) do
```
zachdaniel, line 935: this should be `defp`

zachdaniel (changes requested):

lib/graphql/resolver.ex
dolfinus: I've tried to put `to_resolution(Dataloader.get(...))` into the callback function, but when middleware cannot call it because `to_resolution` is private

lib/ash_graphql.ex
dolfinus: Done

Issue #25, opened 2021-08-03
Support relay for associations


zachdaniel:
I've laid the groundwork for relay support, pushed to master now. There are three remaining steps:

* when fetching `:has_many` or `:many_to_many` relationships, we need to wrap them in a connection if the destination action supports pagination

zachdaniel:
Thanks to the amazing work from @dolfinus, we have the core support for relay in place, so all that is left is to make it work for relationships!

zachdaniel:
I've decided that this is out of scope for the moment. The performance of running that many cursors is highly doubtful. For now, instead, you can simply hide those relationships and/or create a calculation that produces a matching type.

dolfinus:
> The performance of running that many cursors is highly doubtful.

What do you mean?

zachdaniel:
Well, its not so much that there would be issues with performance above and beyond the obvious, but if you make one request that has a bunch of related cursors, to seek one cursor you'd need to make the *entire* request again, refetching all cursors. So its a more scalable approach to do that by using a non-nested field for each relationship you want to page through. (i.e use a filter on the `read<associatedResource>` query)

dolfinus:
But the same issue is possible with offset pagination, isn't it?

zachdaniel:
Yeah, but I think people are less likely to use offset pagination on large datasets. I didn't mean to imply that we should *never* add relay for associations, just that I wasn't going to invest more in doing it now because it would be "better" to prefer top-level pagination. If you're interested in working on relay for associations (will have some difficult design considerations, but should be doable) I'm happy to reopen this and we could have that conversation.

PR #36, opened 2022-05-28
improvement: Implement GraphQL Relay support

dolfinus:
Implement connection object (edge = page_info + nodes) with keyset pagination, compatible with GraphQL Relay.

Solves items 2 and 3 of #25. Item 1 requires more investigation.

Issues to be discussed:
1. Keyset pagination in Ash is not consistent with Relay specification. Specification [requires](https://relay.dev/graphql/connections.htm#sec-Edge-order) `When before: cursor is used, the edge closest to cursor must come last in the result edges`, but Ash returns records from the beginning of the page instead of end. I've marked this behavior as a bug in test cases.
2. Keyset pagination in Ash is currently designed to fetch only next pages, not previous ones. So the way of how `hasPreviousPage` is being calculated may be quite ugly.
3. I'm not sure if `count` field should be present in the `pageInfo` object. It does not return the number of all records in the datalayer, and also it does not return the total number of pages. Instead it returns the number of records matching the keyset condition.
4. If someone is passing invalid cursor value, the exception `(ArgumentError) incorrect padding` is raised and printed to logs, not to the error message returned to user

### Contributor checklist


zachdaniel:
Wow, this is great work! I've skimmed most of your code and it looks excellent.

1. Could we just reverse the results after running the query to handle this case?
2. The counterpart to `more?` for previous pages is not necessarily `less?`, it would be something like `more_previous?`, but that isn't so easy to calculate...would we need to run the same query, swapping `before` and `after` and adding a `limit 1` to check to see if there is a previous page? I haven't thought about keyset pagination in a while, so I'll have to think on this one.
3. `count` *should* always map to the number of records you would get if you traverse all the pages and get all the records, meaning that any filters applied to the request should also be applied to the count. Is this not what you're seeing?
4. where is this raised? If we can make that raise a better error, then we can implement `AshGraphql.Error` for that better exception

This is great work, I'm sure many people will find it very useful!

dolfinus:
1. According to the same [Relay specification](https://relay.dev/graphql/connections.htm#sec-Edge-order): `The ordering of edges should be the same when using first/after as when using last/before, all other arguments being equal. It should not be reversed when using last/before`. So if `sort` is reversed to make a proper query to the datalayer, the result should be reversed back to normal order.
2. Swapping `after` and `before` arguments will cause incorrect result. For example, there are records "a", "b" and "c", you've passed arguments `(first: 1, after: "b.cursor")`, resulting records will be filtered by `id > b.id`. If you change `after` to `before`, filter will look like `id < b.id`, so record "b" will not pass it.
3. `if you traverse all the pages and get all the records`. For example, there are records "a", "b" and "c", you've passed arguments `first: 2, after: "b.cursor")`, `count` will be 1, because there is only one record with `id > b.id`. But you can traverse pages backwards using `last` + `before` to get 2 other records, which were not counted by previous call. Is that OK?
4. https://github.com/dolfinus/ash_graphql/blob/0699a3a2872d3e77b2e421a5bf7acc9e12b5ad7c/test/relay_test.exs#L846
```
13:18:43.281 [warning] `f24054f1-9666-46ef-a5bf-ef75ded2b023`: AshGraphql.Error not implemented for error:

** (Ash.Error.Unknown) 

* ** (ArgumentError) incorrect padding
    (ash 1.52.0-rc.9) lib/ash/engine/request.ex:222: Ash.Engine.Request.next/1
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:530: Ash.Engine.advance_request/1
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:455: Ash.Engine.fully_advance_request/2
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:389: Ash.Engine.do_run_iteration/2
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:354: anonymous fn/2 in Ash.Engine.run_iteration/1
    (elixir 1.13.2) lib/enum.ex:2396: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:349: Ash.Engine.run_iteration/1
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:180: Ash.Engine.run_to_completion/1
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:129: Ash.Engine.do_run/2
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:98: Ash.Engine.run/2
    (ash 1.52.0-rc.9) lib/ash/actions/read.ex:123: Ash.Actions.Read.run/3
    (ash_graphql 0.17.5-rc1) test/support/api.ex:1: AshGraphql.Test.Api.read/2
    (ash_graphql 0.17.5-rc1) lib/graphql/resolver.ex:164: AshGraphql.Graphql.Resolver.resolve/2
    (absinthe 1.7.0) lib/absinthe/phase/document/execution/resolution.ex:230: Absinthe.Phase.Document.Execution.Resolution.reduce_resolution/1
    (absinthe 1.7.0) lib/absinthe/phase/document/execution/resolution.ex:185: Absinthe.Phase.Document.Execution.Resolution.do_resolve_field/3
    (absinthe 1.7.0) lib/absinthe/phase/document/execution/resolution.ex:170: Absinthe.Phase.Document.Execution.Resolution.do_resolve_fields/6
    (absinthe 1.7.0) lib/absinthe/phase/document/execution/resolution.ex:88: Absinthe.Phase.Document.Execution.Resolution.walk_result/5
    (absinthe 1.7.0) lib/absinthe/phase/document/execution/resolution.ex:67: Absinthe.Phase.Document.Execution.Resolution.perform_resolution/3

    (ash 1.52.0-rc.9) lib/ash/engine/request.ex:222: Ash.Engine.Request.next/1
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:530: Ash.Engine.advance_request/1
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:455: Ash.Engine.fully_advance_request/2
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:389: Ash.Engine.do_run_iteration/2
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:354: anonymous fn/2 in Ash.Engine.run_iteration/1
    (elixir 1.13.2) lib/enum.ex:2396: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:349: Ash.Engine.run_iteration/1
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:180: Ash.Engine.run_to_completion/1
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:129: Ash.Engine.do_run/2
    (ash 1.52.0-rc.9) lib/ash/engine/engine.ex:98: Ash.Engine.run/2
    (ash 1.52.0-rc.9) lib/ash/actions/read.ex:123: Ash.Actions.Read.run/3
    (ash_graphql 0.17.5-rc1) test/support/api.ex:1: AshGraphql.Test.Api.read/2
    (ash_graphql 0.17.5-rc1) lib/graphql/resolver.ex:164: AshGraphql.Graphql.Resolver.resolve/2
    (absinthe 1.7.0) lib/absinthe/phase/document/execution/resolution.ex:230: Absinthe.Phase.Document.Execution.Resolution.reduce_resolution/1
    (absinthe 1.7.0) lib/absinthe/phase/document/execution/resolution.ex:185: Absinthe.Phase.Document.Execution.Resolution.do_resolve_field/3
    (absinthe 1.7.0) lib/absinthe/phase/document/execution/resolution.ex:170: Absinthe.Phase.Document.Execution.Resolution.do_resolve_fields/6
    (absinthe 1.7.0) lib/absinthe/phase/document/execution/resolution.ex:88: Absinthe.Phase.Document.Execution.Resolution.walk_result/5
    (absinthe 1.7.0) lib/absinthe/phase/document/execution/resolution.ex:67: Absinthe.Phase.Document.Execution.Resolution.perform_resolution/3
```

zachdaniel:
1. Thinking about it, if this is what its doing in core then we need to reverse it in ash core, not here, so if you don't mind opening an issue for it in ash core that would be great.
2. I was more wondering how we would do this in ash core, to determine a `has_previous_page?`. knowing if there is `has_next_page?` (what we currently call just `more?`) is easy, you just add 1 to the limit and see if you got anything back. But if we want to add `has_previous_page?` to keyset pagination I believe we may need to run an additional query.
3. This is a bug that should also be opened in ash core directly. it sounds like `count` is incorrect for keyset pagination.
4. what a surprisingly unhelpful stacktrace, sorry about that! But the only place I can think of that coming from is here: https://github.com/ash-project/ash/blob/main/lib/ash/page/keyset.ex#L92. In that function we return `{:error, <the raised error>}` but we should return something like `Ash.Error.Query.InvalidKeyset` or something like that (we'd need to add a new error for it).

zachdaniel:
We will also want to address this: https://github.com/ash-project/ash/issues/255 in order for keyset pagination to be able to load all of the same things as limit/offset based pagination.

dolfinus:
I've created a bunch of issues in the Ash repo:
1. https://github.com/ash-project/ash/issues/333
2. https://github.com/ash-project/ash/issues/334
3. https://github.com/ash-project/ash/issues/335
4. https://github.com/ash-project/ash/issues/336
5. https://github.com/ash-project/ash/issues/337

zachdaniel:
Excellent work! Great issues, very descriptive, I'm sure we can get all of this handled :)

dolfinus:
> Would you be interested in taking a stab at it?

Sure, thank you!

dolfinus:
I've updated this PR to match the current state of `main` branch. Could you please review?

One change that was not present in previous version - `count` field was removed from `PageInfo` type to avoid expensive record count calculation. This simply does not match the while idea of Relay protocol

zachdaniel:
So I think we can still have a `count` field, since actions configure wether or not they should be countable. However, I think its fine to leave it out for now. It would be additive, so no big deal.

To refresh my memory here, the last thing we actually need is to fix the keyset pagination in ash core right?

dolfinus:
Yes, the main issue was #333, and few validation issues mentioned in https://github.com/ash-project/ash_graphql/pull/36#issuecomment-1140522003

zachdaniel:
So I think it currently all hinges on this: https://github.com/ash-project/ash/issues/255 Let me take a look and see what is involved here.

zachdaniel:
Well, it looks like it was actually a lot easier than I expected. For now, its not configurable. We just make two queries every time. The other method is more of an optimization anyway. We can add that back in on-demand.

zachdaniel:
Alright, #333 has been taken care of. All that is left, I believe, is this: https://github.com/ash-project/ash/issues/255

zachdaniel:
@dolfinus one other thing we need to do before we ship this: we should put the `relay?: true` toggle at the api, instead of the resource, perhaps?

dolfinus:
It's complicated.

From the one hand, this is more flexible:
```elixir
  graphql do
    type(:relay_tag)

    queries do
      get :get_relay_tag, :read
      list :get_relay_tags, :read_paginated, relay? true
    end
  end
```
than:
```elixir
  graphql do
    type(:relay_tag)

    relay? true

    queries do
      get :get_relay_tag, :read
      list :get_relay_tags, :read_paginated
    end
  end
```
because user can mark that only a specific query implements Relay interface, not all of them.
But if user have many list queries with the same behavior, all of them should be marked, which can be too verbose.

From the other hand, user already pointed that pagination is `keyset?`, so maybe `relay?` is redundant:
```elixir
  actions do
    defaults([:create, :update, :destroy, :read])

    read :read_paginated do
      pagination(required?: true, keyset?: true, countable: true)
    end
  end
```
But it is possible to set both `offset?` and `keyset?` at the same time, which may be ambiguous.
Also user can have different APIs for the same resource, e.g. legacy JSON:API + new GraqphQL, which should be implemented at the same time for backward compatibility, and changing the action leads to changing both APIs instead of just one of them.

So I'm not sure which option is the best one.

zachdaniel:
🤔 Okay, I think its a good point that `relay? true` in a single query is actually a better starting experience. As with everything in Ash, we will just have to roll-forward and add more configuration as people seek out more complex setups. The important part is the behavior.

dolfinus:
So the preferable option is:
```elixir
  graphql do
    type(:relay_tag)

    relay? true

    queries do
      get :get_relay_tag, :read
      list :get_relay_tags, :read_paginated
    end
  end
```
Right?

dolfinus:
Actually, `relay?` in API instead of query may be required by 1 item of https://github.com/ash-project/ash_graphql/issues/25

zachdaniel:
🤔 I think we can look at the main action being called, and use that to determine how we fetch related data.

zachdaniel:
So it should still be okay on the query.

dolfinus:
Then this options should be moved from query to the action

zachdaniel:
Sorry, it should still be on the query. Actions don't know about graphql. The query is what puts the action into the graphql, so to correct my earlier statement: "we can look at the top level query being run, and use that to determine how we fetch related data".

dolfinus:
> The query is what puts the action into the graphql

This is true for top-level queries, but connections can be create without query.

I mean something like:
```graphql
query {
  post(filter: {id: {eq: 1}}) {
    comments(first: 10) {
      text
    }
  }
}
```
`post` here is a `query`, but `comments` is not.

If `relay?` option is set in the query, user should create query for listing comments only to mark `comments` relation as returning Relay object instead of simple offset pagination. Is that right?

zachdaniel:
Potentially. Putting this configuration on the action is not really an option though, either way, as the graphql extension can't add configuration to actions. So my thought was that if an action has `relay? true` then it should cause all child actions that *can* be relay compatible to be relay compatible. But that is a bit janky, you're right. So if we leave `relay? true` on the resource level, then we know which associations to use connections with.

zachdaniel:
I think perhaps we start with `relay? true` at the resource level, as that is the simplest change. We can add richer customization after we confirm that `relay? true` behaves well on its own?

zachdaniel:
Alright, so I've tried this out, and I think this may actually be pretty much good to go. It doesn't support pagination of loaded data, but currently nothing in Ash does. So I think if we merge this, we can get relay support going for top level resources, and then approach making the relationships support it. Thoughts?

dolfinus:
Ok, I moved `relay?` from `api` to `query`, please check.

zachdaniel:
As always, very well done! I'll fix up some stuff in ash core and we can get this merged :D Eventually, we will add things to tell it to use connections for associations, but this is a solid base to build on. Seriously, *very* good work.

dolfinus:
Regarding the behavior of `hasPreviousPage` and `hasNextPage`, I've added a workaround for `more?` with `before` cursor which should fix a bug mentioned in previous review iteration

zachdaniel:

PR #42, opened 2022-10-26
chore(docs): getting started link in readme points to a 404.

jimsynz:


zachdaniel (dismissed):
Really we should link to ash-hq here. I've approved it because you fixed what you set out to fix, but would you mind making that change?

jimsynz:
Done.

PR #45, opened 2022-11-01
Add empty mutations declaration to doc

oco-adam:
Fixes ash-project/ash_graphql/issues/44


zachdaniel:
So this is a bit complicated. I'm going to have to fix this in a different way because this fix only works if you have some resources that define a mutation, and the ergonomics of that aren't fun. I'm exploring options at the moment.

PR #46, opened 2022-11-02
improvement: Add Ash.Error.Changes.InvalidChanges AshGraphql implementation

janajri:
### Contributor checklist

### DIFF
--- a/lib/error.ex
+++ b/lib/error.ex
@@ -2,6 +2,18 @@ defprotocol AshGraphql.Error do
   def to_error(exception)
 end
 
+defimpl AshGraphql.Error, for: Ash.Error.Changes.InvalidChanges do
+  def to_error(error) do
+    %{
+      message: error.message,
+      short_message: error.message,
+      vars: Map.new(error.vars),
+      code: Ash.ErrorKind.code(error),
+      fields: [error.field]
+    }
+  end
+end
+
 defimpl AshGraphql.Error, for: Ash.Error.Query.InvalidQuery do
   def to_error(error) do
     %{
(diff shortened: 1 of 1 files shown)


lib/error.ex
```diff
@@ -2,6 +2,18 @@ defprotocol AshGraphql.Error do
...
 end
 
+defimpl AshGraphql.Error, for: Ash.Error.Changes.InvalidChanges do
+  def to_error(error) do
+    %{
+      message: error.message,
+      short_message: error.message,
+      vars: Map.new(error.vars),
+      code: Ash.ErrorKind.code(error),
+      fields: [error.field]
```
zachdaniel, line 12: I think we want something like `List.wrap(error.field) ++ error.fields` in this case, IIRC that error can have a `field` or a `fields`.

lib/error.ex
```diff
@@ -2,6 +2,18 @@ defprotocol AshGraphql.Error do
   def to_error(exception)
 end
 
+defimpl AshGraphql.Error, for: Ash.Error.Changes.InvalidChanges do
```
janajri, line 5: I noticed UnknownError was also missing -- not sure if these are intentional or not.

lib/error.ex
janajri: I just checked but it looks like it only has fields option
>   def_ash_error([:fields, :message, :validation], class: :invalid)

lib/error.ex
zachdaniel: Yeah, we don't want to automatically surface unknown errors up to the user (because they could contain arbitrary data)

lib/error.ex
zachdaniel: I believe they all have a `field` key by default, but that this error specifically adds an additional `fields` field. Confusing, I know.

zachdaniel: But what you have is fine :) IF you're using this error you're expected to set `fields` and not `field`

zachdaniel: But what you have is fine :) IF you're using this error you're expected to set `fields` and not `field`

PR #59, opened 2023-02-24
fix: always return a list from pkey_fields

barnabasJ:
### Contributor checklist

The function `manage_pkey_fields` that calls `pkey_fields` could not handle a single value being returned as it tries to map over the result.


zachdaniel (approved):

zachdaniel:

PR #62, opened 2023-03-23
improvement: thread context from `PlugHelpers.get_context` through to `Query` and `Changeset`

bcksl:
Blocked on https://github.com/ash-project/ash/pull/534

### Contributor checklist


zachdaniel:
This looks great, but we can't assume that they are using the new stuff, so we need to first check for the new value and then the old value.

zachdaniel (approved):

zachdaniel:
Got some CI stuff to take care of

zachdaniel:
👍 I think we're close to that release. Had some bugs in main but the ones I know of are now resolved

zachdaniel:
This new release has gone out, so if you update this branch to use it we can move forward

zachdaniel:
You'll need to update the `mix.exs` to depend on the newer version of ash.

PR #63, opened 2023-03-27
improvement: pass `description` through to Absinthe in more cases

bcksl:
Check the `TODO:`s for a couple of additional places where adding descriptions would require supporting them in core. Operators maybe not relevant, but `Ash.Resource.Calculation.Argument` probably worthwhile. See https://github.com/ash-project/ash/blob/de943509f73ff0c6ef3749feb3cc6ab6e4d99d2d/lib/ash/resource/calculation/calculation.ex#L103

This is not 100% comprehensive, there are a lot of `InputObjectDefinition`s that could be candidates.

### Contributor checklist

### DIFF
--- a/lib/resource/resource.ex
+++ b/lib/resource/resource.ex
@@ -421,6 +421,7 @@ defmodule AshGraphql.Resource do
           complexity: {AshGraphql.Graphql.Resolver, :query_complexity},
           module: schema,
           name: to_string(query.name),
+          description: Ash.Resource.Info.action(resource, query.action).description,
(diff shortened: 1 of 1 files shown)


zachdaniel (dismissed):
Surprised we don't have descriptions on calculation arguments. You can make it work for now by using `Map.get(argument, :description)` so when we add it we won't have to update this?

jimsynz:
Hi @bcksl please address the credo warnings and I'll make sure that we get this merged.

PR #64, opened 2023-04-03
Add missing definition for `Ash.Type.Time` type

marmor157:
This should not cause any issues as `Absinthe` declares this type in `Absinthe.Type.Custom` same as it is also done for `:datetime`.

Sadly, I've didn't add any tests for it since there were no specific tests for `:datetime` either.


zachdaniel:

PR #66, opened 2023-04-13
improvement: support manual relationships

ahey:
Prior to this patch, an error was raised `* No such attribute [resource]_id`, since the data loader attempted to load the relationship via a related id field, which does not exist for manual relationships.


zachdaniel (approved):

zachdaniel:

ahey:
@zachdaniel thank you for the speedy merge!

PR #68, opened 2023-04-17
feat: create types for maps with constraints

barnabasJ:
### Contributor checklist


lib/ash_graphql.ex
```diff
@@ -398,6 +401,13 @@ defmodule AshGraphql do
         attribute_type = unwrap_type(attribute.type)
 
         case attribute_type do
+          Ash.Type.Map ->
+            if attribute.constraints[:fields] do
```
zachdaniel, line 405: You'll probably need to map over the sub-fields the way that the `union` type does.

lib/resource/resource.ex
```diff
@@ -2334,6 +2336,107 @@ defmodule AshGraphql.Resource do
     String.to_atom(to_string(type) <> "_sort_field")
   end
 
+  def map_definitions(resource, schema, env) do
+    resource = Ash.Type.NewType.subtype_of(resource)
```
zachdaniel, line 2340: I don't think we should need to do this.

lib/resource/resource.ex
```diff
@@ -2334,6 +2336,107 @@ defmodule AshGraphql.Resource do
...
+            end
+          else
+            nil
+          end
+
+        [
+          type_name
+        ]
+        |> Enum.filter(& &1)
+        |> Enum.map(fn type_name ->
```
zachdaniel, line 2392: We need to do this differently for input and non-input, since some types have different input representations.

zachdaniel (dismissed):
Looking great! Got some changes to make, and then ideally we can support this in nested cases.

barnabasJ:
Nested maps should now work too.

zachdaniel (approved):

PR #69, opened 2023-04-27
feat: add `filterable_fields` to limit generated filters

bcksl:
Adds the ability to specify which fields a filter should be made available for:

```elixir
graphql do
  filterable_fields [:name, :age]
end
```

### Contributor checklist

### DIFF
--- a/lib/resource/resource.ex
+++ b/lib/resource/resource.ex
@@ -303,6 +303,12 @@ defmodule AshGraphql.Resource do
         doc:
           "Whether or not to create the GraphQL object, this allows you to manually create the GraphQL object.",
         default: true
+      ],
+      filterable_fields: [
+        type: {:list, :atom},
+        required: false,
+        doc:
+          "A list of fields that are allowed to be filtered on. Defaults to all filterable fields for which a GraphQL type can be created."
       ]
     ],
     sections: [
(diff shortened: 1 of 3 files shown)


lib/resource/resource.ex
```diff
@@ -2164,10 +2170,16 @@ defmodule AshGraphql.Resource do
...
+    all_fields =
+      boolean_filter_fields(resource, schema) ++
+        attribute_filter_fields(resource, schema) ++
+        relationship_filter_fields(resource, schema) ++
+        aggregate_filter_fields(resource, schema) ++ calculation_filter_fields(resource, schema)
+
+    filterable_fields = AshGraphql.Resource.Info.filterable_fields(resource)
+
+    all_fields
+    |> Enum.filter(&(is_nil(filterable_fields) or &1.identifier in filterable_fields))
```
zachdaniel, line 2182: unfortunately, we can't actually do this using the identifier, because it may have been remapped using the `field_names` option. So we'll want to do it in the individual functions.

zachdaniel (changes requested):

lib/resource/resource.ex
```diff
@@ -2172,9 +2178,11 @@ defmodule AshGraphql.Resource do
 
   defp attribute_filter_fields(resource, schema) do
     field_names = AshGraphql.Resource.Info.field_names(resource)
+    filterable_fields = AshGraphql.Resource.Info.filterable_fields(resource)
 
     resource
     |> Ash.Resource.Info.public_attributes()
+    |> Enum.filter(&(is_nil(filterable_fields) or &1.name in filterable_fields))
     |> Enum.filter(&filterable?(&1, resource))
```
zachdaniel, line 2186: Alright, this is very close :D

Two things:
1. pull main into your branch, some stuff has changed here
2. `Enum.filter() |> Enum.filter()` will have credo yell at you, you'll need to combine them.

zachdaniel (approved):
Looks great 🥳

PR #70, opened 2023-04-28
Document how to query for errors in a mutation

moxley:
### Contributor checklist

- [x] Docs

### DIFF
--- a/documentation/topics/graphql-generation.md
+++ b/documentation/topics/graphql-generation.md
@@ -230,11 +230,22 @@ mutation($input: CreateTicketInput!) {
       subject
       status
     }
+    errors {
+      code
+      fields
+      message
+      shortMessage
+      vars
+    }
   }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


documentation/topics/graphql-generation.md
```diff
@@ -230,11 +230,22 @@ mutation($input: CreateTicketInput!) {
...
   }
 }
 ```
 
-Notice that the resulting ticket data is wrapped in a `result` object.
+**Note**
+
+- The resulting ticket data is wrapped in AshGraphql's `result` object.
+- Validation errors are wrapped in an `errors` object, also specified in the query.
+  AshGraphql does this instead of exposing errors in GraphQL's standard `errors` array.
```
zachdaniel, line 248: If we're highlighting it here, we might want to also highlight that this behavior is just the default. It can be changed with:
```elixir
use Ash.Api, extensions: [AshGraphql.Api]

graphql do
  root_level_errors? true
end
```

zachdaniel (approved):

PR #83, opened 2023-07-13
update field policy test

barnabasJ:
I updated the test to match my expectations. The assertion didn't do anything before. I changed the attribute on the resource to be non-nullable to show the problem with masking in the case of GraphQL.


zachdaniel:
I think some code got lost somewhere 😢 Merging, will need to figure this out.

PR #91, opened 2023-08-28
feat: add more sort order values

barnabasJ:


zachdaniel:
I like this, but I think we might want to expose it as `NULL` instead of `NIL` in the graphql layer.

lib/api/api.ex
```diff
@@ -200,15 +200,15 @@ defmodule AshGraphql.Api do
           module: schema,
           identifier: :desc_nils_first,
           __reference__: AshGraphql.Resource.ref(env),
-          name: "DESC_NILS_FIRST",
-          value: :desc
+          name: "DESC_NULLS_FIRST",
+          value: :dialyzer_succ_typings
```
zachdaniel: This looks like an accident 😆

barnabasJ: I blame copilot :rofl:

zachdaniel:

PR #92, opened 2023-08-29
test: add failing test for an aggregate of a calculation

ahey:
An error occurs at compile time when an aggregate of a calculation is defined:

```
== Compilation error in file test/support/schema.ex == ** (MatchError) no match of right hand side value: {:error, "Must provide field type for max"}
    lib/resource/resource.ex:2278: AshGraphql.Resource.filterable?/2
    (elixir 1.15.4) lib/enum.ex:4265: Enum.filter_list/2
    (elixir 1.15.4) lib/enum.ex:4266: Enum.filter_list/2
    lib/resource/resource.ex:2229: AshGraphql.Resource.aggregate_filter_fields/2
    lib/resource/resource.ex:2195: AshGraphql.Resource.resource_filter_fields/2
    lib/resource/resource.ex:1159: AshGraphql.Resource.args/5
    lib/resource/resource.ex:425: anonymous fn/6 in AshGraphql.Resource.queries/5
    (elixir 1.15.4) lib/enum.ex:1693: Enum."-map/2-lists^map/1-1-"/2
```

### DIFF
--- a/test/read_test.exs
+++ b/test/read_test.exs
@@ -594,6 +594,44 @@ defmodule AshGraphql.ReadTest do
     assert %{data: %{"getCompositePrimaryKey" => %{"id" => ^id}}} = result
   end
 
+  test "aggregate of a calculation" do
+    post =
+      AshGraphql.Test.Post
+      |> Ash.Changeset.for_create(:create,
+        text: "foo",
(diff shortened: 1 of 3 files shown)


zachdaniel:

PR #95, opened 2023-10-03
refactor: make the node in relay edges non-null

barnabasJ:
I had to do some front-end work today and felt like this would be better. The spec doesn't say if it has to be nullable or non-nullable. Same with the Edge, I don't think it would make sense to have null values as edges. 

I couldn't think of a situation where we would actually return null for a node, pls correct me if I'm wrong.


zachdaniel (approved):

zachdaniel:
Oh, wait I think you've used list instead of non null 😂

barnabasJ:
> Oh, wait I think you've used list instead of non null 😂

Oh, yeah, copy paste 🤦‍♂️ will fix

zachdaniel:

PR #100, opened 2023-11-24
fix: make sure generic actions are authorized

barnabasJ:
### Contributor checklist

### DIFF
--- a/lib/graphql/resolver.ex
+++ b/lib/graphql/resolver.ex
@@ -41,11 +41,17 @@ defmodule AshGraphql.Graphql.Resolver do
               :gql_query,
               query_name,
               metadata do
+          opts = [
+            actor: Map.get(context, :actor),
+            authorize?: AshGraphql.Api.Info.authorize?(api),
+            tenant: Map.get(context, :tenant)
+          ]
+
           result =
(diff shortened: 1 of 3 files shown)


lib/graphql/resolver.ex
```diff
@@ -41,11 +41,17 @@ defmodule AshGraphql.Graphql.Resolver do
...
               metadata do
+          opts = [
+            actor: Map.get(context, :actor),
+            authorize?: AshGraphql.Api.Info.authorize?(api),
+            tenant: Map.get(context, :tenant)
+          ]
+
           result =
             %Ash.ActionInput{api: api, resource: resource}
             |> Ash.ActionInput.set_context(get_context(context))
```
barnabasJ, line 52: I'm a bit confused about this because it seems like `context` is already the `context`, and those values are never attached to the action. 

This looks for another nested `context` or `ash_context` inside the context. This is also done in all the other resolvers, and it seems like the actor is only set in the context because of the opts. Not sure if I'm missing something there?

barnabasJ: I think everything is working for now. I just wanted to make sure I understand this.

zachdaniel:

PR #106, opened 2024-01-18
Relay refetching support

rbino:
Add support for Relay global IDs and the `node` query, see https://relay.dev/graphql/objectidentification.htm

### Contributor checklist

### DIFF
--- a/lib/graphql/resolver.ex
+++ b/lib/graphql/resolver.ex
@@ -136,7 +136,7 @@ defmodule AshGraphql.Graphql.Resolver do
            action: action,
            identity: identity,
            modify_resolution: modify
-         } = gql_query}
+         } = gql_query, relay_ids?}
       ) do
     case handle_arguments(resource, action, arguments) do
       {:ok, arguments} ->
@@ -165,7 +165,7 @@ defmodule AshGraphql.Graphql.Resolver do
             tenant: Map.get(context, :tenant)
           ]
 
-          filter = identity_filter(identity, resource, arguments)
+          filter = identity_filter(identity, resource, arguments, relay_ids?)
 
           query =
             resource
@@ -279,7 +279,7 @@ defmodule AshGraphql.Graphql.Resolver do
         %{arguments: args, context: context} = resolution,
         {api, resource,
          %{name: query_name, type: :read_one, action: action, modify_resolution: modify} =
-           gql_query}
+           gql_query, _relay_ids?}
       ) do
     metadata = %{
       api: api,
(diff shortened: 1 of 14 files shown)


rbino:
Opened this as draft to collect some early feedback, global IDs are already working, I will tackle the `node` query in the next days

lib/resource/resource.ex
```diff
@@ -417,6 +425,27 @@ defmodule AshGraphql.Resource do
     end
   end
 
+  def encode_relay_id(%resource{} = record) do
+    primary_key = encode_primary_key(record)
+
+    "#{resource}:#{primary_key}"
```
zachdaniel, line 431: We should use the graphql type name as the prefix here, as opposed to the resource.

rbino:
@zachdaniel I think we're almost there _but_ right now only the ID gets selected because it seems the fields contained in the type-specific fragments don't get included in the selected fields by Ash. Is there something where, e.g., interfaces are implemented that I can take as reference to see if I can make this work?

lib/graphql/resolver.ex
```diff
@@ -2536,6 +2543,33 @@ defmodule AshGraphql.Graphql.Resolver do
...
+      # TODO: what if there's no get query? It should probably not implement the node interface
+      query = get_query(resource)
+
+      # We pass relay_ids? as false since we pass the already decoded primary key
+      put_in(resolution.arguments.id, primary_key)
+      |> resolve({api, resource, query, false})
+    else
+      _ ->
+        # TODO: is this return value ok?
+        Absinthe.Resolution.put_result(resolution, {:error, "Invalid primary key"})
```
zachdaniel, line 2558: Not sure what the return value should be, probably an `Ash.Error.Framework.AssumptionFailed`. Although we could avoid the case entirely by requiring that each resource has a `get` query at compile time.

rbino:
@zachdaniel this should be now good to go, I'm now only implementing the node interface for resources with a primary key get.

I also had to change some of the code regarding field selection for the reason I mentioned above (fields coming from fragment weren't being selected), it is now working correctly.

lib/resource/resource.ex
```diff
@@ -4118,4 +4154,11 @@ defmodule AshGraphql.Resource do
       name
     end
   end
+
+  def primary_key_get_query(resource) do
```
zachdaniel, line 4158: I think one of the things we'll need to do soon here is make this configurable, but the strategy of choosing an existing gql query for it to mimic is pretty neat.

zachdaniel:
looks like some credo failures to address, then we can merge it :)

rbino:
It was complaining about the two `TODO`s I had left, removed them

zachdaniel:
😆 of course. We can also just delete that credo rule.

zachdaniel:
awesome work! 🚀 Thank you for your contribution! 🚀

PR #107, opened 2024-01-25
improvement: return a map from decode_relay_id/1

rbino:
Slight API improvement over #106.

This makes it more ergonomic to partially match on the return value (and it also makes it more explicit by explicitly labeling the two parts).

Also add tests for relay id encoding/decoding.

### Contributor checklist


zachdaniel:

PR #109, opened 2024-01-29
feat: add Relay ID translation in mutation and queries

rbino:
Adds a new option for queries and mutations that defines which arguments or attributes will use a global Relay ID and their type. This allows automatically decoding them before hitting their action.

### Contributor checklist

### DIFF
--- a/documentation/topics/relay.md
+++ b/documentation/topics/relay.md
@@ -25,3 +25,34 @@ use AshGraphql, relay_ids?: true
 ```
 
 This allows refetching a node using the `node` query and passing its global ID.
+
+### Translating Relay Global IDs passed as arguments
+
+When `relay_ids?: true` is passed, users of the API will have access only to the global IDs, so they
+will also pass use them when an ID is required as argument. You actions, though, internally use the
+normal IDs defined by the data layer.
+
+To handle the translation between the two ID domains, you can use the `relay_id_translations`
+option. With this, you can define a list of arguments that will be translated from Relay global IDs
+to internal IDs.
+
+For example, if you have a `Post` resource with an action to create a post associated with an
+author:
+
+```elixir
+create :create do
+  argument :author_id, :uuid
+
+  # Do stuff with author_id
+end
+```
+
+You can add this to the mutation connected to that action:
+
+```elixir
+mutations do
+  create :create_post, :create do
+    relay_id_translations [input: [author_id: :user]]
+  end
+end
+```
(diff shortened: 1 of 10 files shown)


rbino:
@zachdaniel the feature is working, but it's missing some error handling and some more niceties. Opening as draft to collect feedback on the direction/naming/etc.

zachdaniel:
I'm wondering if there is a way for us to figure this out automatically for cases with `manage_relationship`?

rbino:
I think this would be orthogonal to `managed_relationships`, from what I understand, the main purpose of `managed_relationships` is generating input types from nested relationships, so I think it's mainly used for nested resource creation (e.g. `create_post_with_tags`) but less so with relationships that don't create the related resource (e.g. `create_post_with_author`, which doesn't create the author). Also, an ID could also be passed in a non-standard argument (like a map) and handled with a manual change, which wouldn't be detectable.

The high level idea I have in mind to see if this can be automatically generated is more or less this:
- Do something similar to `managed_relationship` and retrieve all changes in the action which use a `ManageRelationship` change
- From these changes, filter the ones which map the argument to a primary key of the related resource
- Emit the translation middleware for those arguments (also taking care of the nesting, e.g. the wrapping `input` object)

Given we would still need the manual way to do this to provide an escape hatch for cases not covered by the autogeneration, I don't know if you prefer fixing the manual way and then working on the autogeneration in a separate PR or start directly with the autogeneration.

zachdaniel:
Yes, great points. We need both 👍 Doesn't really matter which one we start with 😄

zachdaniel:
As for managed relationship type generation, you *should* be able to tap into our existing logic for generating types for managed relationships

rbino:
@zachdaniel ready for review, this implements the manual part as discussed above. I will then open an issue to track the implementation of the automatically derived translations.

rbino:
@zachdaniel rebased to fix the conflicts in the cheatsheet. I've also shortened the docs in the DSL and linked to the Relay guide instead as suggested by the Spark warning.

documentation/topics/relay.md
```diff
@@ -25,3 +25,34 @@ use AshGraphql, relay_ids?: true
 ```
 
 This allows refetching a node using the `node` query and passing its global ID.
+
+### Translating Relay Global IDs passed as arguments
+
+When `relay_ids?: true` is passed, users of the API will have access only to the global IDs, so they
+will also pass use them when an ID is required as argument. You actions, though, internally use the
```
zachdaniel, line 32:
```suggestion
will also need to use them when an ID is required as argument. You actions, though, internally use the
```

zachdaniel:

Issue #105, opened 2024-01-07
Make some standard fields non nullable


ahey:
Generating a graphql layer for my front-end in typescript, I noticed some fields are unnecessarily nullable. To achieve a more pleasant typescript experience, I propose the following fields be made `non null`:

- `errors` as returned by a mutation. In case of no errors, an empty list will be returned. Furthermore, each item in the error list can also be made not null, since it either exists in the list or not, and we don't want nulls in the list.
- `error.fields`

In addition to these fields, I would also like to see the the `_result` fields be made non nullable, as they will always contain the `result` and `errors` fields, except in the case of `root_level_errors?` being enabled, in which case i suppose the `_result` field should be nullable.

[See here for an example of the changes](https://github.com/ash-project/ash_graphql/compare/main...ahey:ash_graphql:non-null-lists?expand=1) (not ready for merge, just for discussion)

This would be a breaking change to the existing api. We could keep the existing behaviour as default, and add a config setting to enable the new behaviour. The existing behaviour should be considered deprecated and we could warn to use the flag, and default to use it in v1.0.0.

What do you think?

zachdaniel:
IIRC `result` will be `null` if `errors` is present, and `errors` will be `null` if the `result` is present, right? However, items in the error list being non null can definitely be done.

rbino:
I think (also from the diff) that @ahey was referring to the fact that the Result type _as a whole_ is non-nullable (as you said you always either have a `result` or some `errors`), so

```
createFoo(input: CreateFooInput): CreateFooResult
```

should instead be

```
createFoo(input: CreateFooInput): CreateFooResult!
```

This (as mentioned above) is particularly useful when integrating with TypeScript where nullability is important for type generation.

Another doubt I have is: should the input also be non-nullable?

zachdaniel:
Ah, yes you are right, we can definitely make that return type non nullable. Going from nullable -> not nullable on a returned type isn't breaking, so I'd definitely accept a PR on that front. As for non-nullable input, that one is more tricky. Not all actions require some kind of input, and if all keys are not required, the input could also be not required. It would be a breaking change to require it, even if we only did it in cases where there was at least one required input inside of it, because nullable -> not nullable on an input type is theoretically breaking. Although in those scenarios the action would fail 100% of the time... 🤷 so maybe fine to just make both changes. overall result type not nullable, and input not nullable *only if* at least one member is not nullable.

barnabasJ:
[...]

PR #110, opened 2024-01-31
First round of improvements around non-nullability

rbino:
First round of implementation of what was discussed in #105 

This PR makes result types non-nullable.

### DIFF
--- a/lib/resource/resource.ex
+++ b/lib/resource/resource.ex
@@ -599,7 +599,9 @@ defmodule AshGraphql.Resource do
             module: schema,
             name: to_string(mutation.name),
             description: Ash.Resource.Info.action(resource, mutation.action).description,
-            type: String.to_atom("#{mutation.name}_result"),
+            type: %Absinthe.Blueprint.TypeReference.NonNull{
+              of_type: String.to_atom("#{mutation.name}_result")
+            },
             __reference__: ref(__ENV__)
           }
         end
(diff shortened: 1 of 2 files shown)


zachdaniel:
For the ids in updates being non nullable, that would be a breaking change for any queries people have, right? I think we'll have to put that behind a configuration for now, and undo it when we do 1.0 for this library.

rbino:
Wouldn't the same thing that was discussed in #105 regarding making the input non-nullable apply (see below)?

> It would be a breaking change to require it, even if we only did it in cases where there was at least one required input inside of it, because nullable -> not nullable on an input type is theoretically breaking. Although in those scenarios the action would fail 100% of the time... 🤷 so maybe fine to just make both changes.

Anyway, adding a config works for me too, but which kind of config (i.e. at what level) should it be?

zachdaniel:
Yeah, but the changes you made in the test made me realize that it's not enough that it wouldn't have worked, because it will break people's type declarations. i.e if they have `id: ID` `GraphQL` will yell at them because its not an optional value anymore. We can use application config, there should be some other examples in this repo. I'll cut a 1.0 around the time we do 3.0 of Ash (1-2 months)

rbino:
@zachdaniel I've left only the non-breaking result change in this PR and I'll do a separate one with IDs and input objects behind an application config

zachdaniel (approved):

PR #111, opened 2024-01-31
improvement: make mutation arguments non-null

rbino:
As discussed in #105 and #110, put this behind an opt-in configuration to avoid breaking existing code.
The ID in update mutations is always non-null if non-null mutation arguments are allowed, while input is non-null if it's allowed _and_ there is at least a non-null field in the input.

Document the newly added config variable in the getting started guide.

### Contributor checklist


zachdaniel (approved):

zachdaniel:

PR #113, opened 2024-02-07
Data layer functions

barnabasJ:
Allow DataLayer Filters (like, ilike from postgres in filter) 

depends on 
* https://github.com/ash-project/ash_postgres/pull/205
* https://github.com/ash-project/ash/pull/874

Very naive first implementation. It works in my project already, but I feel I'm probably missing some edge cases.
Somehow, the filter is already only applied to String attributes, but I'm not completely sure why.

### DIFF
--- a/lib/resource/resource.ex
+++ b/lib/resource/resource.ex
@@ -2106,6 +2106,7 @@ defmodule AshGraphql.Resource do
 
     fields =
       Ash.Filter.builtin_operators()
+      |> Enum.concat(Ash.DataLayer.functions(resource))
       |> Enum.filter(& &1.predicate?)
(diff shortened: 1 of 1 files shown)


zachdaniel:
(Thats why it knows what types to support)

zachdaniel:
Okay, maybe not 😆 I’m not sure how it could only be supporting strings. I think the function variation will need to do something similar to what the operator code does. In fact, ideally they’d share the same code instead of branching.

zachdaniel (approved):

barnabasJ:
> Is this good to go?

Everything is fine from my side.

zachdaniel:

PR #114, opened 2024-02-09
fix: make results nullable again if root level errors are enabled

rbino:
Building upon #110, this restores the old behaviour of the result being nullable when root level errors are present.

While the result is guaranteed to not be nullable in standard conditions (since either result or errors are always present), when errors are moved to the root level it could become null, so declaring it non-nullable propagates the null up to the data field.

This actually causes compatibility problems with some client libraries (e.g. Relay) that expect the inner result to be null, _not_ data, if there's an error.

This also adds dedicated RootLevelErrors versions of the Api and the Schema since the configuration is accessed at compile time now, so put_env was not enough to test them correctly.

### Contributor checklist

### DIFF
--- a/test/errors_test.exs
+++ b/test/errors_test.exs
@@ -12,8 +12,6 @@ defmodule AshGraphql.ErrorsTest do
   end
 
   test "errors can be configured to be shown in the root" do
-    Application.put_env(:ash_graphql, AshGraphql.Test.Api, graphql: [root_level_errors?: true])
-
     resp =
       """
       mutation CreatePost($input: CreatePostInput) {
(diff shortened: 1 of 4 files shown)


test/errors_test.exs
```diff
@@ -38,7 +36,7 @@ defmodule AshGraphql.ErrorsTest do
 
     assert {:ok, result} = resp
 
-    assert %{data: nil, errors: [%{message: message}]} = result
+    assert %{data: %{"createPost" => nil}, errors: [%{message: message}]} = result
```
rbino, line 39: Note that this change returns the behaviour to the one pre #110, so it's not breaking (there was no release between the two)

zachdaniel:

PR #120, opened 2024-03-27
improvement: add `show_fields` option as explicit fields whitelist

vonagam:
There is `hide_fields` option. How about adding `show_fields` option that if present acts like an explicit list of allowed fields?

My use case is that I want to move graphql section out of resources into separate folder with `Spark.Dsl.Fragment`. But it means that looking just at that file you won't have context about present fields. With this you would have some idea.

Plus it is good from security perspective since you might add an attribute and forget to add it to `hide_fields`.


zachdaniel:
I see no problem with it 👍

zachdaniel:

PR #121, opened 2024-03-28
feat: allow resources without types

### LINKED ISSUE #119: Type should be optional for resources that only expose generic actions
I am working on a resource that only needs to expose a single generic action. In that case, since the resource itself is never exposed in GraphQL, the `type` option is actually not needed, but AshGraphql requires me to provide one.

To avoid breaking compatiblity and ensure that the choice is informed I think there should be an additional option (e.g. `generate_type? false`) and either `type` or `generate_type? false` has to be provided.

Alternatively, `type` could be enforced only for resources that expose some non-generic action.

The main alternative right now is moving the generic action in another resource that already has a connected type.

```elixir
  graphql do
    generate_type? false
  end
```

rbino:
Ensure they only expose generic action queries. Add checks to ensure that either `type :resource_type` or `generate_object? false` is passed if it's needed.

Close #119

### Contributor checklist


zachdaniel:
This looks good to me, but one thing I want to be sure of is that it doesn't break cases where users have relationships to non-graphql resources. The current behavior is that those relationships are ignored. Is that the same behavior still?

rbino:
Yep, [here](https://github.com/ash-project/ash_graphql/pull/121/files#diff-d7c91b14f9704acb8a2d158976636282e55f408496f5f764671a60f9faddb24eR99) it still checks if `AshGraphql.Resource` is in the extensions. I've added a non-graphql resource in tests to verify it correctly compiles (if I add `extensions: [AshGraphql.Resource]` it fails to compile, exactly like before when `type` was required)

zachdaniel:

PR #123, opened 2024-03-30
improvement!: port AshGraphql to Ash 3.0

rbino:
## Steps

- Step 1: update Ash

- Step 2: mass rename Api to Domain

- Step 3: Ash.Query.expr -> Ash.Expr.expr

Also change ref interpolation

- Step 4: remove all warnings

- Step 5: remove registries from tests

- Step 6: fix filter

- Step 7: private? -> !public?

- Step 8: Ash.Calculation -> Ash.Resource.Calculation

- Step 9: use depend_on_resources/1 -> resources/1

- Step 10: add Domain to all resources

- Step 11: use Ash module for all actions

- Step 12: add public? true all around

- Step 13: remove verbose? from options passed during Domain calls

- Step 14: add simple_sat

- Step 15: Ash.ErrorKind is no more, so remove code from errors

- Step 16: sprinkle default_accept :* around tests

- Step 17: replace Ash.Changeset.new/2 with Ash.Changeset.for_*

- Step 18: calculation fixups

    - Context is now a struct and arguments go under the arguments key
    - Function based calculations receive a list of records
    - Add a select to query-based loads
    - select -> load

- Step 19: pass the correct name to pass the policy in tests

- Step 20: Ash.Query.new/2 is no more

- Step 21: add AshGraphql.Resource.embedded? utility function

Use that instead of Ash.Type.embedded_type?(resource_or_type) since resources are not types anymore

- Step 22: handle struct + instance_of: Resource in unions

Resources are not type anymore so they need to be passed this way in unions

- Step 23: ensure we only check GraphQL actions for pagination

All reads are now paginated by default, so this triggered a compilation error

- Step 24: swap arguments for sort on calculations
  
- Step 25: remove unused debug? option

### Contributor checklist


rbino:
This is the branch with one commit per step: https://github.com/rbino/ash_graphql/tree/3.0-step-by-step

test/support/resources/channel/calculations/messages_calculation.ex
```diff
@@ -1,20 +1,24 @@
...
+    offset = context.arguments.offset || 0
 
     [
       :channel_message_count,
-      messages: AshGraphql.Test.Message |> Ash.Query.limit(limit) |> Ash.Query.offset(offset)
+      messages:
+        AshGraphql.Test.Message
+        |> Ash.Query.limit(limit)
+        |> Ash.Query.offset(offset)
+        |> Ash.Query.select([:type, :text])
```
rbino, line 15: I had to explicitly add a `select` here, otherwise the load would not load the fields. Not sure if this is normal or if this could be a bug in Ash.

lib/error.ex
```diff
@@ -8,7 +8,6 @@ defimpl AshGraphql.Error, for: Ash.Error.Changes.InvalidChanges do
       message: error.message,
       short_message: error.message,
       vars: Map.new(error.vars),
-      code: Ash.ErrorKind.code(error),
```
rbino, line 11: This probably will break existing code which expects a `code` inside error, is there anything else we can use insead of `Ash.ErrorKind.code`?

test/support/resources/channel/calculations/messages_calculation.ex
rbino: Now that I think about it maybe what is missing here is `Ash.Query.new()`

rbino: Ok it's normal
https://hexdocs.pm/ash/3.0.0-rc.5/upgrade.html#calculation-loads-do-not-select-all-related-fields-by-default

lib/error.ex
zachdaniel: Yeah, so unfortunately what has to be done here for backwards compatibility is for the original error codes to be hand-copied here. Basically responsibility for generating this additional metadata has been moved out of core and into the api extensions. So this one would be something like `"invalid_changes"`, and so-on.

PR #138, opened 2024-04-29
feat: allow passing custom descriptions to queries and mutations

### LINKED ISSUE #137: Allow adding a description at the query/mutation level
Right now, the description shown in the Graphql docs is taken from the action description, but there could be multiple queries/mutation which actually use the same action (e.g. `:list` and `:get` could both use the default `:read` action).

There should be a way to customize description at the query/mutation level.

It's possible to work around this by having multiple different actions with different descriptions which do the same thing, but this quickly spirals out of control if the actions need to be kept in sync.

```elixir
queries do
  get :get_post, :read do
    description "Gets a single post by its ID"
  end

  list :list_posts, :read do
    description "Lists all posts"
  end
end
```

The description would basically be

```
query.description || query_action.description
```

rbino:
Close #137

### Contributor checklist

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -13,6 +13,7 @@ spark_locals_without_parens = [
   depth_limit: 1,
   derive_filter?: 1,
   derive_sort?: 1,
+  description: 1,
   destroy: 2,
   destroy: 3,
   encode_primary_key?: 1,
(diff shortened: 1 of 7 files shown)


zachdaniel (approved):

zachdaniel:

PR #144, opened 2024-05-03
improvement: make mutation result errors list non-nullable

ahey:
This change makes the errors list of mutation results non-nullable. The type now looks like `errors: [MutationError!]!` instead of `errors: [MutationError!]`.

### Contributor checklist


rbino:
How does that behave when `root_level_errors` is defined? My doubt is that in that case, the `errors` field inside the mutation will not be an empty list but `nil` (which would error during resolution at the Absinthe level if someone queries the inner `errors` field)

zachdaniel:
Yeah, I think we need to only do this if `root_level_errors?` is `false`.

ahey:
> Yeah, I think we need to only do this if root_level_errors? is false.

I have looked into this further, and my understanding is that it's ok even when root_level_errors is true.

> How does that behave when root_level_errors is defined? My doubt is that in that case, the errors field inside the mutation will not be an empty list but nil (which would error during resolution at the Absinthe level if someone queries the inner errors field)

[I have now added a test case to cover this](https://github.com/ahey/ash_graphql/blob/non-nullable-mututation-errors/test/errors_test.exs#L69). When `root_level_errors` is enabled, the root level errors field is not returned (as per existing behaviour and graphql spec), and the errors field inside the mutation result is returned as an empty list.

zachdaniel (approved):

zachdaniel:

PR #162, opened 2024-05-23
fix: don't deduplicate argument types by argument name

rbino:
As mentioned in the upgrade guide, the point of dropping auto generation of types was to avoid conflicts when two arguments had the same name but different types. Now that we require explicit generation, we can actually support that usecase for, e.g., arguments with the same name but different types in create/update mutations.

### Contributor checklist

### DIFF
--- a/lib/resource/resource.ex
+++ b/lib/resource/resource.ex
@@ -3176,7 +3176,6 @@ defmodule AshGraphql.Resource do
           !Enum.empty?(Ash.Type.NewType.constraints(&1.type, &1.constraints)[:fields] || []) &&
           define_type?(&1.type, &1.constraints))
     )
-    |> Enum.uniq_by(& &1.name)
   end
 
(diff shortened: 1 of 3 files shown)


rbino:
Drafted to check if we have to deduplicate for type instead, now that we removed deduplication for name

zachdaniel:
Looks great! Merge when ready 😁

zachdaniel:
Oh, interesting...I think we deduplicate in type name somewhere already? There are definitely already cases where people reuse the same enum/new type and it doesn't cause issues.

rbino:
I'm hitting this: if I define a NewType with both `graphql_type` and `graphql_input_type` (with different types) and I use both (the first as a calculation output type, the second as argument input), Absinthe complains that the type is duplicate.

I'm checking to see if I can reproduce this in a test case.

rbino:
So the duplication thing is actually orthogonal to this (I'm opening an issue about it), so this can be merged

PR #166, opened 2024-05-25
feat: relationship pagination

### LINKED ISSUE #131: Allow paginating nested relationships
Currently, passing `relay? true` to a list query enables Connection pagination on the query result. Sometimes though the requirement is having a _field_ paginated with a Connection. See for example  the `versions` field in the [Package](https://docs.github.com/en/graphql/reference/objects#package) type in the Github GraphQL API.

There should be a way to require a list field to be paginated.

Note that while `relay? true` is a property of _the query_, the paginated field should actually be a property of _the type_, because if e.g. an `Author` has a `posts` field that returns a `PostConnection`, then everything that returns an `Author` will have to consume `posts` with a `Connection`. This also means that the root query could be non paginated while the field is (which is actually a usecase I'm actively interested in).

So the syntax could be something like:

```elixir
graphql do
  type :author

  relay_connections: [:posts]

  queries do
    list :authors_paginated, :read, relay?: true
    list :authors, :read, paginate_with: nil # Will still have :posts paginated with the Connection
  end
end
```

I guess the only lists that could be paginated with `Connection` are relationships (but I might be wrong)

rbino:
Resources can now define how their relationships are paginated. This is a property of the Graphql type of the resource and it allows arbitrarily nested paginated relationships.

The PR also contains a fix to the existing pagination, which was not working correctly when a keyset strategy was used with an action that also supported offset pagination.

Close #131 

### Contributor checklist

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -42,6 +42,7 @@ spark_locals_without_parens = [
   metadata_names: 1,
   metadata_types: 1,
   modify_resolution: 1,
+  paginate_relationship_with: 1,
   paginate_with: 1,
   primary_key_delimiter: 1,
   read_action: 1,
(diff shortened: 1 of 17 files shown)


zachdaniel:

PR #177, opened 2024-06-11
chore: add tests on paginated relationships

davidebriani:
Following on ash-project/ash#1234, this PR adds tests to verify that relationships can be queried on the results of GraphQL queries and mutations.

Querying paginated relationships on the result of deletions doesn't correctly work yet.
Indeed, Ash currently queries relationships using a lateral join but after the resource deletion has happened, so it looks like nothing is related.

### Contributor checklist

### DIFF
--- a/test/relationship_pagination_test.exs
+++ b/test/relationship_pagination_test.exs
@@ -223,43 +223,45 @@ defmodule AshGraphql.RelationshipPaginationTest do
     assert [%{"name" => "Award 2"} | _] = results
   end
 
-  test "works when nested" do
-    movie =
-      AshGraphql.Test.Movie
-      |> Ash.Changeset.for_create(:create, title: "Movie")
-      |> Ash.create!()
+  describe "works when nested" do
+    test "on return values for queries" do
+      movie =
+        AshGraphql.Test.Movie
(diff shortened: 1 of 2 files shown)


zachdaniel:
The "looks like nothing is related" bit is pretty complex. We can try to "simulate" this in `ash` which is that when relationships are loaded on a query for deletion we could just run the query first, save the relationship data, and then delete without the load. That way we could give a best effort of what the relationships look like. If you think about what its asking its actually kind of ambiguous anyway though, like are you asking for "what it looks like afterwards"? If so, the destroy actions should probably not return anything at all TBH. Anyway, that can be its own conversation :)

zachdaniel:
@davidebriani are the tests meant to be failing? Is this PR demonstrating a bug?

zachdaniel:
Or does it just need latest fixes from `ash`?

davidebriani:
@zachdaniel the tests should work with the latest version of Ash.
I assume you'd want to tag an Ash release first, update deps here, and finally I'll rebase the PR

zachdaniel:
I guess either way, merging it should be fine. The subprojects test in `ash` will tell us.

zachdaniel:

PR #178, opened 2024-06-11
add resource query to action struct

barnabasJ:
### Contributor checklist


zachdaniel:
🤔 is there a case where we are getting an action but no resource somewhere? We'll need to populate this in a transformer if we want to use it.

zachdaniel:
oh, I didn't realize the repo this was in. Whoops.

zachdaniel:

PR #186, opened 2024-06-28
Add (failing) test for partial results of forbidden field query

skanderm:
Updates the forbidden field test to include public attributes (currently failing)


zachdaniel:
Thanks for bringing this to my attention! The problem was the field nullability here, and I've added to the authorization guide for how to address it, and added a small tool to simplify the addressing of it `nullable_fields [:foo, :bar, :baz]`

PR #187, opened 2024-07-01
fix: use checked constraints

barnabasJ:
I had this fail because I created a new Type (subtype of map with a field that was an enum)

### Contributor checklist

### DIFF
--- a/lib/resource/resource.ex
+++ b/lib/resource/resource.ex
@@ -4160,7 +4160,7 @@ defmodule AshGraphql.Resource do
         else
           if Spark.implements_behaviour?(type, Ash.Type.Enum) do
             if function_exported?(type, :graphql_type, 1) do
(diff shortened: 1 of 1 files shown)


lib/resource/resource.ex
```diff
@@ -4160,7 +4160,7 @@ defmodule AshGraphql.Resource do
         else
           if Spark.implements_behaviour?(type, Ash.Type.Enum) do
             if function_exported?(type, :graphql_type, 1) do
-              type.graphql_type(attribute.constraints)
+              type.graphql_type(constraints)
```
barnabasJ, line 4163: On line 4121, we check if the attribute has constraints. Otherwise, we use the constraints passed as function arguments. They are used everywhere else in this function except here.

zachdaniel (approved):

PR #205, opened 2024-08-03
chore: Set up GraphQL schema file in the web module namespace

sevenseacat:
This keeps consistency with AshJsonApi, which generates its router in the web module namespace as well

### Contributor checklist


zachdaniel:
There has been some debate as to whether or not the GraphQL schema truly belongs in web, but ultimately this is what absinthe's guides suggest, and so that is what we should do as well. Not to mention the fact that 99% of people use GraphQL only as a web interface.

zachdaniel:

PR #217, opened 2024-10-02
improvement: implement a subscription notification batcher

zachdaniel:
### Contributor checklist

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -11,6 +11,10 @@ config :logger, level: :info
 
 config :ash_graphql, :subscriptions, true
 
+if Mix.env() == :test do
+  config :ash_graphql, :simulate_subscription_slowness?, true
+end
+
 if Mix.env() == :dev do
   config :git_ops,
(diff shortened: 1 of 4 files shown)


barnabasJ:
@zachdaniel, should we merge this now? I can create a PR against main this week, removing the unnecessary data we copy.

barnabasJ:
> Yep! Just need to fix conflicts

I can do that, if you want me to.

zachdaniel:
That would be great, thank you 🙏 you're the best 😍

barnabasJ (approved):

barnabasJ:
@zachdaniel I rebased. However, I had to upgrade Ash because there was an error in the alias test. This also led to an igniter update. That now triggered some Igniter Dialyzer warnings.

It looks like those are just deprecation warnings. So I think it should be ok to merge this?

zachdaniel:
Yep, I'll address those, thanks!

PR #220, opened 2024-10-03
Subscription batcher finish

barnabasJ:
### Contributor checklist

### DIFF
--- a/lib/graphql/resolver.ex
+++ b/lib/graphql/resolver.ex
@@ -796,7 +796,7 @@ defmodule AshGraphql.Graphql.Resolver do
   end
 
   defp do_refetch(query, primary_key_matches) do
-    Ash.read(Ash.Query.do_filter(query, or: primary_key_matches))
+    Ash.read(Ash.Query.do_filter(query, and: primary_key_matches))
   end
 
   defp subcription_field_from_action_type(:create), do: "created"
--- a/lib/graphql/resolver.ex
+++ b/lib/graphql/resolver.ex
@@ -639,7 +639,7 @@ defmodule AshGraphql.Graphql.Resolver do
                 end
               else
                 Enum.map(notifications, fn notification ->
-                  %{type => AshGraphql.Resource.encode_id(notification.data, false)}
+                  %{key => AshGraphql.Resource.encode_id(notification.data, false)}
                 end)
               end
             end)
--- a/lib/graphql/resolver.ex
+++ b/lib/graphql/resolver.ex
@@ -796,7 +796,7 @@ defmodule AshGraphql.Graphql.Resolver do
   end
 
   defp do_refetch(query, primary_key_matches) do
-    Ash.read(Ash.Query.do_filter(query, and: primary_key_matches))
+    Ash.read(Ash.Query.do_filter(query, or: primary_key_matches))
   end
 
   defp subcription_field_from_action_type(:create), do: "created"
--- a/lib/subscription/batcher.ex
+++ b/lib/subscription/batcher.ex
@@ -134,12 +134,9 @@ defmodule AshGraphql.Subscription.Batcher do
     # the tasks processed the batch, so we need to remove only
     # the notifications that where sent
     batch =
-      %{
-        batch
-        | notifications: Enum.reject(batch.notifications, &(&1 in notifications)),
-          count: batch.count - length(notifications),
-          task: nil
-      }
+      batch
+      |> Batch.remove(notifications)
+      |> Map.put(:task, nil)
 
     state =
       %{
--- a/lib/subscription/batcher.ex
+++ b/lib/subscription/batcher.ex
@@ -3,17 +3,6 @@ defmodule AshGraphql.Subscription.Batcher do
   If started as a GenServer, this module will batch notifications and send them in bulk.
   Otherwise, it will send them immediately.
 
-  The Batcher can be configured during compile time using the following settings:
-
-  ```elixir
-  config :ash_graphql, :subscription_batch_settings,
-    # if there are more than 100 `async_limit` notifications, we will start to backpressure
-    async_limit: 100,
-    # if there are less then 50 `send_immediately_threshold` notifications, we will send them immediately
-    send_immediately_threshold: 50,
-    # the interval in milliseconds the batcher waits for new notifications before sending them
-    subscription_batch_interval: 1000
-  ```
   """
   use GenServer
 
@@ -78,6 +67,18 @@ defmodule AshGraphql.Subscription.Batcher do
       do_send(topic, [notification], pubsub, key_strategy, doc)
   end
 
+  @doc """
+  Config options
+
+  `async_limit` (default 100):
+   if there are more than `async_limit` notifications, we will start to backpressure
+
+  `send_immediately_threshold` (default 50): 
+   if there are less then `send_immediately_threshold` notifications, we will send them immediately
+
+  `subscription_batch_interval` (default 1000):
+   the interval in milliseconds the batcher waits for new notifications before sending them
+  """
   def init(config) do
     {:ok,
      %__MODULE__{
--- a/lib/subscription/batcher.ex
+++ b/lib/subscription/batcher.ex
@@ -2,7 +2,6 @@ defmodule AshGraphql.Subscription.Batcher do
   use GenServer
 
   alias Absinthe.Pipeline.BatchResolver
-  alias AshGraphql.Subscription.Runner
 
   require Logger
   @compile {:inline, simulate_slowness: 0}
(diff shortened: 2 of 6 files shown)


zachdaniel:
One more thing we need to do here action is make sure that the notifier doesn't raise or exit when the Bacher it's not alive

barnabasJ:
> One more thing we need to do here action is make sure that the notifier doesn't raise or exit when the Bacher it's not alive

@zachdaniel What do you want to do if it is not started?

barnabasJ:
> Hmm... probably send it synchronously? It's a choice, we should either do nothing or send it synchronously. Maybe even make that configurable? But for now let's send synchronously.

Alright. I wasn't sure if you wanted to "force" people to use the batcher or not.

lib/subscription/runner.ex
```diff
@@ -5,12 +5,54 @@ defmodule AshGraphql.Subscription.Runner do
   Mostly a copy of https://github.com/absinthe-graphql/absinthe/blob/3d0823bd71c2ebb94357a5588c723e053de8c66a/lib/absinthe/subscription/local.ex#L40
   but this lets us decide if we want to send the data to the client or not in certain error cases
   """
+  alias Absinthe.Pipeline.BatchResolver
 
   require Logger
 
   def run_docset(pubsub, docs_and_topics, notification) do
+    batch? = not is_nil(Process.whereis(AshGraphql.Subscription.Batcher))
```
zachdaniel, line 13: I think what we need to do is something like this inside of `Batcher.publish`

```elixir
  def publish(topic, notification, pubsub, key_strategy, doc) do
    case GenServer.call(
           __MODULE__,
           {:publish, topic, notification, pubsub, key_strategy, doc},
           :infinity
         ) do
      :handled ->
        :ok

      :backpressure_sync ->
        do_send(topic, [notification], pubsub, key_strategy, doc)
    end
    catch
      :exit, {:noproc, {GenServer, :call, [__MODULE__, _, _]}} ->
        do_send(topic, [notification], pubsub, key_strategy, doc)
  end
```

lib/subscription/batcher.ex
```diff
@@ -31,16 +31,20 @@ defmodule AshGraphql.Subscription.Batcher do
...
-           __MODULE__,
-           {:publish, topic, notification, pubsub, key_strategy, doc},
-           :infinity
-         ) do
-      :handled ->
-        :ok
-
-      :backpressure_sync ->
-        do_send(topic, [notification], pubsub, key_strategy, doc)
+    if is_nil(Process.whereis(__MODULE__)) do
```
barnabasJ, line 34: I tried it with `catch` but that didn't seem to work.

zachdaniel: 🤔 in what way? Ideally we wouldn't lookup the process twice (one to check and one to send) and then we also have the scenario where it's up when we check but crashes when we call it. The catch version *should* work.

barnabasJ: If what I read is correct, it is because it is not raising an error but sending an exit signal instead. 

This is what Claude came up with, but this felt even more complex to me

```elixir
defmodule SafeCall do
  def call(pid, message, timeout \\ 5000) do
    ref = Process.monitor(pid)
    send(pid, {:$gen_call, {self(), ref}, message})

    receive do
      {:DOWN, ^ref, :process, ^pid, reason} ->
        {:error, reason}
      {^ref, reply} ->
        Process.demonitor(ref, [:flush])
        {:ok, reply}
    after
      timeout ->
        Process.demonitor(ref, [:flush])
        {:error, :timeout}
    end
  end
end

# Usage
case SafeCall.call(pid, :some_message) do
  {:ok, reply} ->
    # Process the reply
  {:error, reason} ->
    # Handle the error (e.g., :noproc, :normal, :killed, etc.)
end
```

lib/subscription/batcher.ex
```diff
@@ -1,14 +1,37 @@
...
   @compile {:inline, simulate_slowness: 0}
 
-  defstruct batches: %{}, total_count: 0, async_limit: 100, send_immediately_threshold: 50
+  settings = Application.compile_env(:ash_graphql, :subscription_batch_settings, [])
+
+  defstruct batches: %{},
+            total_count: 0,
+            async_limit: Keyword.get(settings, :async_limit, 100),
+            send_immediately_threshold: Keyword.get(settings, :send_immediately_threshold, 50),
+            subscription_batch_interval: Keyword.get(settings, :subscription_batch_interval, 1000)
```
zachdaniel, line 31: `subscription_batch_interval` should be fetched from the `opts` on `init` as well. The recommended way to configure things like GenServers is through initialization options instead of compile time options. We can have both if we like, but the runtime configuration is generally better.

The way that it fetches the opts in `init` right now makes this work

```elixir
{Batcher, async_limit: 100}
```

And if we want them to configure it in config files, they should configure the module,

i.e `config :ash_graphql, AshGraphql.Subscription.Batcher, ...`

lib/subscription/batcher.ex
barnabasJ: I'm just talking about the `monitor`/`demonitor` stuff. Not the whole thing

zachdaniel: exits can be caught 🤔 

This works for me:

```elixir
try do
  GenServer.call(:not_existing, :hello)
catch
  :exit, {:noproc, _} -> "no process"
end
```

zachdaniel: exits from *linked processes* cannot be caught, but the exit from `exit/1` can be, which is what this is.

lib/subscription/batcher.ex
barnabasJ: Make sense. I'll remove the compile time again. You can add those options in `application.ex` then. 

But it's ok to also add the `subscription_batch_interval` there? That `TODO` comment was what made me doing this in the first place.

lib/subscription/batcher.ex
barnabasJ: Strange, I added it just like this. But the test still failed with the same error. I'll try again.

barnabasJ: Ok, my bad. I just tested with a "catch all" pattern like

```
catch
  err -> dbg(err0
```

to try and see exactly how the thing would look like, but in that case it wouldn't catch it.

lib/subscription/batcher.ex
zachdaniel: Yeah, we should have that too 👍 I just forgot 😆 Thank you for catching that

lib/subscription/batcher.ex
barnabasJ: Oh, ok. because there are 2 params. `exit` and `{:noproc, _}` 

Is there a way to capture everything regardless of number of parameters. I thought that catch would always just have patterns with one parameter. 

TIL :sweat_smile:

zachdaniel: I think in this case the only thing we want to capture is the noproc, but you can just add
```elixir
catch
_ -> ...something was thrown
_, _ -> ...exit
```

lib/subscription/batcher.ex
```diff
@@ -91,21 +126,46 @@ defmodule AshGraphql.Subscription.Batcher do
...
+    batch = state.batches[topic]
 
-      _ ->
-        :ok
-    end
+    # remove the notifications that where sent from the batch
+    # it is possible that new notifications where added while
+    # the tasks processed the batch, so we need to remove only
+    # the notifications that where sent
+    batch =
```
zachdaniel, line 136: This would be a very expensive comparison. I think we need to figure something else out here, like move the batch to its own key, like "sending batches" that are keyed by their task. Comparing the list of notifications means a cross-wise comparison of every single value inside of every nested map/list etc.

barnabasJ: In that case can we just remove the batch when we start the task? If new notifications would be added it would just create a new batch. 
Right now, we don't do anything with the task that is saved to the batch except check if there is one when we send_all_batches and do nothing in that case. But that would be the same behaviour as removing the batch and not saving the task.

zachdaniel: Yeah, thats a good idea. We just have to *not* decrement the count until the batch has finished.

barnabasJ: I'll do that tomorrow.

lib/graphql/resolver.ex
```diff
@@ -796,7 +796,7 @@ defmodule AshGraphql.Graphql.Resolver do
   end
 
   defp do_refetch(query, primary_key_matches) do
-    Ash.read(Ash.Query.do_filter(query, or: primary_key_matches))
+    Ash.read(Ash.Query.do_filter(query, and: primary_key_matches))
```
zachdaniel, line 799: why the change to and? were refetching a list of records by primary key.

zachdaniel (approved):

zachdaniel:
meant to request changes on the or/and thing. one last thing we should do (in a different PR) is clean the notifications before sending to the batches. like keep only the keys we need, so we do as little copying as possible. IIRC we don't use anything but the data and the action.

lib/graphql/resolver.ex
barnabasJ: Ok, I had to reread the docs of `Map.take` now it makes more sense.

barnabasJ: Yeah, it took me a bit, but I reread some code and docs. The filter is automatically applied as an `and`, and the rest is an or of all the pk matches.

PR #224, opened 2024-10-07
fix: dyalizer and igniter deprecations

barnabasJ:
Dialyzer did have a problem with the batcher code. I just haven't used dialyzer before, so i looked at the wrong ouptut.


zachdaniel:

PR #232, opened 2024-10-30
improvement: use Module.safe_concat

27medkamal:
Use the safer version which raises an error on startup if the module doesn't exist

https://hexdocs.pm/elixir/Module.html#safe_concat/1

### Contributor checklist


zachdaniel:
I think this incurs a compile time dependency which is exactly what we're trying to avoid by using `Module.concat`.

zachdaniel:
Thank you for your contribution, but I believe this change would be problematic for the described reason.

27medkamal:
Right, that makes sense

PR #238, opened 2024-11-12
improvement: Support generic actions without a return type.

jimsynz:
Infer that their return type is `:boolean` and they always return `true`. We considered returning `null` but that's also returned when the query/mutation fails and might convey the wrong information.

### Contributor checklist


jimsynz:
@sevenseacat relevant to your interests.

sevenseacat:
This makes the most sense to me. 

Well `null` for "no return" would make more sense, but then its not distinguishable from the error case, so this is a good compromise!

barnabasJ (approved):
we also used true for mutations without a result in the last 2 companies I worked at, so I'd guess it's not that uncommon as a pattern

PR #246, opened 2024-11-24
fix: load relationships and calculations in fragments

rbino:
Similar to what was done in 2886e0c8c1aea3c943d for selections, we have to do the same thing for loads to make them work with interface fragments (including the Relay node interface).

### Contributor checklist


zachdaniel:

PR #253, opened 2024-12-23
feat: error handling in resources

lusergit:
similarly to error handling in domains, we can allow the `error_handler` option also in resources, so that only errors coming from the specific resource can be customized

### Contributor checklist


zachdaniel:
Something to keep in mind here is that if you have an action that calls other actions, like using `managed_relationship` etc., the errors would still be handled by the primary resource that called the action. I think there is value in this pattern, but that would be a pretty big caveat to the functionality that should be explained in the docs for the option.

lusergit:
Sorry for the late reply. In the last commit i added the caveat you wrote about in the documentation and  slightly updated the logic of error resolution. The documentation should be up to date with the logic. Additionally the CI should pass now :smiley:

zachdaniel (approved):

zachdaniel:

PR #266, opened 2025-01-29
feat: subscription installer

barnabasJ:
### Contributor checklist

depends on https://github.com/ash-project/igniter/pull/239


lib/igniter.ex
```diff
@@ -183,57 +184,158 @@ if Code.ensure_loaded?(Igniter) do
       end
     end
 
+    def setup_application(igniter) do
```
barnabasJ, line 187: I created these two functions because it felt like there was a pattern. Looking at it afterwards, this all feels pretty phoenix specific with `Pheonix.Socket` and `Endpoint`s. 

I would refactor this and make these `defps` that are called as part of `setup_phoenix`. WDYT?

barnabasJ:
As subscriptions are opt in at the moment, we should also make it a flag for the installer to add it?

barnabasJ:
@zachdaniel could you give this an quick glance please and let me know if I should continue this way?

zachdaniel:
Do you think subscriptions could just be unflagged now? I feel like no one uses it because it's behind a flag 😆. Lets just open it up and install this automatically.

barnabasJ:
> Some tests would be good to add, specifically for the complex case of adding the parser

The parser part is actually not new :sweat_smile: but I agree. I'll look into adding some installer tests in general.

barnabasJ:
> Do you think subscriptions could just be unflagged now? I feel like no one uses it because it's behind a flag 😆. Lets just open it up and install this automatically.

@zachdaniel I feel pretty good about it after the experiment I made last week. Only thing to consider is that we should maybe rename the [pubsub](https://github.com/ash-project/ash_graphql/blob/71de915581366980ad624d439221deeb88414e7e/lib/resource/resource.ex#L283) option in the subscription dsl. Because the way Absinthe subscriptions are [called](https://github.com/absinthe-graphql/absinthe/blob/3d0823bd71c2ebb94357a5588c723e053de8c66a/lib/absinthe/subscription.ex#L90) you actually need to configure the Endpoint here and not the PubSub directly?

At least it tripped me up, when I implemented subscriptions last week. 

Not sure I'm happy with just calling it enpoint either though?

zachdaniel:
I think that module is anything with a broadcast function right?

barnabasJ:
> I think that module is anything with a broadcast function right?

It's something that implements the [Absinthe.Subscription.Pubsub](https://github.com/absinthe-graphql/absinthe_phoenix/blob/39bdf98e1e45d5f264b29043a9d1244681e3f826/lib/absinthe/phoenix/endpoint.ex#L4) so i guess pubsub is probably the right word. It just tripped me up because I tried to configure the `Phoenix.PubSub` and that didn't work.

zachdaniel (dismissed):

PR #273, opened 2025-02-10
feat: Type and Query complexity callbacks

### LINKED ISSUE #270: Feat: Complexity callback in DSL
I'd like to set the complexity of an action exposed through Ash GraphQL, similarly to when using Absinthe directly:

Example from the [Absinthe Docs](https://hexdocs.pm/absinthe/1.7.8/complexity-analysis.html#complexity-analysis-1)

```elixir
  query do
    field :people, list_of(:person) do
      arg :limit, :integer, default_value: 10
      complexity fn %{limit: limit}, child_complexity ->
        # set complexity based on maximum number of items in the list and
        # complexity of a child.
        limit * child_complexity
      end
    end
  end
```

For my use case, I'd like to increase the complexity of a sign-in mutation to prevent multiple sign-in attempts in a single request. Adding a callback to the DSL similar to absinthe would be ideal:

```elixir
  graphql do
    type :my_user

    hide_fields([:id, :hashed_password])

    queries do
      get :sign_in, :sign_in_with_password do
        as_mutation?(true)
        complexity(fn _args, child_complexity -> child_complexity + @some_large_complexity_value end)
      end
    end
  end
```

mbuhot:
Resolves #270 

A custom complexity callback can be provided for a type to be used when that type is resolved through a relationship.

A query-level complexity callback applies when the resource has a top-level query in the schema.

### Contributor checklist


test/support/resources/sponsored_comment.ex
mbuhot, line 1: @zachdaniel I found I needed to have a type-level complexity callback for use when the schema is being resolved through a relationship, rather than a direct query.

Should we keep both, or just use the type-level complexity callback for all reads if one exists?

zachdaniel (commented):
Looks great! I think we should make these MFAs instead of just mod/fun pairs (hardcoded args get prepended to our arguments). That allows better reuse, i.e `complexity {MyApp.Complexity, :complexity, [__MODULE__, :very_complex]}` etc.

test/support/resources/sponsored_comment.ex
zachdaniel: Both is good.

mbuhot:
@zachdaniel I'm having some trouble forwarding the MFA through to Absinthe:

```
 2) test custom complexity calculation (AshGraphql.ReadTest)
     test/read_test.exs:424
     ** (Absinthe.AnalysisError) Invalid value returned from complexity analyzer.

     Analyzing field:

       sponsoredComments

     Defined at:

       /Users/michaelbuhot/src/ash-project/ash_graphql/lib/resource/resource.ex:4338

     Got value:

         {AshGraphql.Test.SponsoredComment, :query_complexity, []}

     The complexity value must be a non negative integer.
```

It looks like only functions, {mod, fun} tuples and constants are supported:

https://github.com/absinthe-graphql/absinthe/blob/main/lib/absinthe/phase/document/complexity/analysis.ex#L107-L129

zachdaniel:
Ah, right you are. I had a similar complication at some point in the past and did some super hacky shit to get around it. In that case, never mind 😄

zachdaniel:
Please run `mix docs` and `mix spark.formatter` and `mix docs` and commit the result.

PR #286, opened 2025-03-12
Properly non_null() inner items in list when constraint is supplied for arrays in NewType of :map

olivermt:
Fixed two things:
- Pulled constraint from right place (was a bug where it tried to pull constraints from the outer constraints and not from each individual new attribute)
- Utilize the maybe null wrapper and default to not wrapping null unless explicitly supplying nil items

Open question: Maybe we **should** actually default to `nil_items?: false` (as in not use my defensive default to `true`) so that we get the same default setting as for attributes?

It might be confusing that array types does not produce non_null inner items for self defined NewType maps.
Alternatively, that is a bug that needs to be fixed in Ash.
If so then maybe we should make a bug in Ash and reference that bug from this comment I put in?

### DIFF
--- a/lib/resource/resource.ex
+++ b/lib/resource/resource.ex
@@ -4639,7 +4639,14 @@ defmodule AshGraphql.Resource do
   end
 
   defp do_field_type({:array, type}, nil, resource, input?, constraints) do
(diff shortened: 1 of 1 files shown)


zachdaniel (approved):

zachdaniel:

PR #289, opened 2025-03-17
Test making a newtype with a resource as a struct inside

olivermt:
Tests that map types can be made with resources inside

### DIFF
--- a/test/support/resources/comment.ex
+++ b/test/support/resources/comment.ex
@@ -11,6 +11,7 @@ defmodule AshGraphql.Test.Comment do
 
     queries do
       list :list_comments, :read
+      action :list_ranked_comments, :ranked_comments
     end
 
     mutations do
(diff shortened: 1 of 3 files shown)


test/support/resources/comment.ex
```diff
@@ -35,6 +35,22 @@ defmodule AshGraphql.Test.Comment do
     read :paginated do
       pagination(required?: true, offset?: true, countable: true)
     end
+
+    action :ranked_comments, {:array, RankedComment} do
+      run(fn _input, _ctx ->
+        res =
+          Ash.read!(__MODULE__)
```
olivermt, line 42: I didn't get as far as seeing if I had to do some fixturing before this will actually return something, since it won't compile, so I have no idea if this will work or not. But I think the whole test is a feasible and nice thing to have tested.

zachdaniel:

PR #301, opened 2025-04-14
fix: handle invalid query error different formats

artiom:
Follow-up to https://github.com/ash-project/ash/pull/1960. Handle both `field` and `fields` in `InvalidQuery` error.

### DIFF
--- a/lib/error.ex
+++ b/lib/error.ex
@@ -16,12 +16,7 @@ end
 
 defimpl AshGraphql.Error, for: Ash.Error.Query.InvalidQuery do
   def to_error(error) do
-    error_fields =
-      if is_nil(error.field) do
-        Map.get(error, :fields, [])
-      else
-        [error.field]
-      end
+    fields = List.wrap(error.field || Map.get(error, :fields) || [])
 
     %{
       message: error.message,
--- a/lib/error.ex
+++ b/lib/error.ex
@@ -16,12 +16,19 @@ end
 
 defimpl AshGraphql.Error, for: Ash.Error.Query.InvalidQuery do
   def to_error(error) do
+    error_fields =
+      if is_nil(error.field) do
+        Map.get(error, :fields, [])
+      else
+        [error.field]
+      end
+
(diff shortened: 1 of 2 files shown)


lib/error.ex
```diff
@@ -16,12 +16,19 @@ end
 
 defimpl AshGraphql.Error, for: Ash.Error.Query.InvalidQuery do
   def to_error(error) do
+    error_fields =
+      if is_nil(error.field) do
+        error.fields
```
zachdaniel, line 21: This also needs to use `Map.get(error, :fields, [])`

artiom: Done.

lib/error.ex
```diff
@@ -16,12 +16,19 @@ end
 
 defimpl AshGraphql.Error, for: Ash.Error.Query.InvalidQuery do
   def to_error(error) do
+    error_fields =
+      if is_nil(error.field) do
+        Map.get(error, :fields, [])
+      else
+        [error.field]
+      end
```
vonagam, line 24:
```suggestion
    fields = List.wrap(error.field || Map.get(error, :fields) || [])
```

There is a small difference here `Map.get(error, :fields) || []` vs `Map.get(error, :fields, [])` - second option would return `[]` for older Ash as there is no `fields` field, but will return (as I assume) nil for newer versions as there will be such key.

artiom: Thanks for the suggestion 👍

lib/error.ex
```diff
@@ -16,12 +16,14 @@ end
...
   def to_error(error) do
+    fields = List.wrap(error.field || Map.get(error, :fields) || [])
+
     %{
       message: error.message,
       short_message: error.message,
       vars: Map.new(error.vars),
       code: "invalid_query",
-      fields: [error.field]
+      fields: error_fields
```
vonagam, line 26:
```suggestion
      fields: fields
```

vonagam (approved):

PR #307, opened 2025-05-04
Support for returning relay encoded id when subscribing to destroy events

### LINKED ISSUE #306: Subscriptions on :destroy action's don't respect relay_ids
**Describe the bug**
If I have AshGraphql setup with `relay_id?: true`, it seems that if I also setup a subscription for a destroy action type, the data received in the subscription has the primary key of the resource, and not the relay id.

I think this is due to the `resolve/2` function [here](https://github.com/ash-project/ash_graphql/blob/8e0771dde59d4f4b8c703016f0c6ff62d3e9b23a/lib/graphql/resolver.ex#L711). If we look further down on line 834, it seems that the call to encode the id has a hardcoded false as the second param, which tells it not to use relay style id, which can conflict with what the user setup for relay ids.

```elixir
notification.action_type in [:destroy] ->
  resolution
  |> Absinthe.Resolution.put_result(
    {:ok,
      %{
        String.to_existing_atom(
          subcription_field_from_action_type(notification.action_type)
        ) => AshGraphql.Resource.encode_id(notification.data, false)
      }}
  )

```

**To Reproduce**
A minimal set of resource definitions and calls that can reproduce the bug.

```elixir
graphql do
    type :my_resource

    subscriptions do
      pubsub MyApp.Endpoint

      subscribe :resource_added do
        action_types :create
      end

      subscribe :resource_updated do
        action_types :update
      end

      subscribe :resource_deleted do
        action_types :destroy
      end
    end
  end
```

The subscriptions for all of the above return the proper relay id, except for the `:resource_deleted`.

**Expected behavior**
If the user has `relay_ids?: true`, then the subscription messages for deletion should return the relay id, and not the plain primary key of the resource.

\*\* Runtime

- Elixir version: 1.18.3
- Erlang version: OTP 27
- OS
Ash version: 3.5.8
- any related extension versions
Ash Graphql: 1.7.9

**Additional context**

mike1o1:
Closes #306 

Currently, if attempting to subscribe to a destroy event, the message that is received has the shape of something like:

```json
"myResource": {
  "destroyed": "{myUUID}"
}
```

However, if a user has a schema with `relay_ids?: true`, this is the wrong type of ID, as it doesn't get encoded as a relay id.

This PR addresses that by passing the `relay_ids?` value down through the domains and resources, and eventually to the `resolve/2` function which gets added to the Absinthe middleware. When encoding the id for a `:destroy` action, the `relay_ids?` setting will now be respected.

---

#### Tests
I couldn't find a way to change the `relay_ids?` value at runtime during the test, so this unfortunately requires setting up a whole new `RelaySchema` and associated `RelayDomain` and `RelaySubscribable` resource. While verbose, this seemed like the most straight forward way to keep existing tests and schema in place.

### Contributor checklist

### DIFF
--- a/lib/ash_graphql.ex
+++ b/lib/ash_graphql.ex
@@ -249,7 +249,8 @@ defmodule AshGraphql do
                 all_domains,
                 unquote(resources),
                 action_middleware,
-                unquote(schema)
+                unquote(schema),
+                unquote(relay_ids?)
               )
               |> Enum.reduce(blueprint_with_mutations, fn subscription, blueprint ->
                 Absinthe.Blueprint.add_field(blueprint, "RootSubscriptionType", subscription)
--- a/lib/domain/domain.ex
+++ b/lib/domain/domain.ex
@@ -251,7 +251,7 @@ defmodule AshGraphql.Domain do
     )
   end
 
-  def subscriptions(domain, all_domains, resources, action_middleware, schema) do
+  def subscriptions(domain, all_domains, resources, action_middleware, schema, relay_ids?) do
     resources
     |> Enum.filter(fn resource ->
       AshGraphql.Resource in Spark.extensions(resource)
@@ -262,7 +262,8 @@ defmodule AshGraphql.Domain do
         all_domains,
         &1,
         action_middleware,
-        schema
+        schema,
+        relay_ids?
       )
     )
   end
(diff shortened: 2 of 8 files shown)


zachdaniel:
Looks like some credo errors to fix.

zachdaniel:

PR #310, opened 2025-05-06
fix: define shared mutations options

vonagam:
It is mostly documentation thing but also there is a small fix - `modify_resolution` was defined specifically only for `get`, `create` and `action`, not for other queries/mutations, even thought the code handles presence of `modify_resolution` in all action types.


zachdaniel:

PR #311, opened 2025-05-07
fix: refactor internal `AshGraphql.Resource.mutation_fields/5`

vonagam:
It is a refactor but, again, there is a fix included.

- The fix (a little bit breakish): when getting an override type for an argument - [the override name](https://github.com/ash-project/ash_graphql/blob/e7c267a7b07d9bf4e89fb905ab99b002fe07b2f9/lib/resource/resource.ex#L1488) (if present) was used instead of original. I think original should be used, as, for example, it [gets used](https://github.com/ash-project/ash_graphql/blob/e7c267a7b07d9bf4e89fb905ab99b002fe07b2f9/lib/resource/resource.ex#L4720-L4724) for attribute type overrides.

- There was inconsistency in what is used for argument field identifier - original name or override - looking at codebase it seems that it should be the original.

- There was an [unreachable code path](https://github.com/ash-project/ash_graphql/blob/e7c267a7b07d9bf4e89fb905ab99b002fe07b2f9/lib/resource/resource.ex#L1537-L1551) - a truthy `find_manage_change` result guarantees that [`managed_relationship`](https://github.com/ash-project/ash_graphql/blob/e7c267a7b07d9bf4e89fb905ab99b002fe07b2f9/lib/resource/resource.ex#L1582) returns a truthy value.

### DIFF
--- a/lib/resource/resource.ex
+++ b/lib/resource/resource.ex
@@ -1432,7 +1432,8 @@ defmodule AshGraphql.Resource do
 
   defp mutation_fields(resource, schema, action, type, hide_inputs \\ []) do
     field_names = AshGraphql.Resource.Info.field_names(resource)
-    argument_names = AshGraphql.Resource.Info.argument_names(resource)
+    argument_names = AshGraphql.Resource.Info.argument_names(resource)[action.name]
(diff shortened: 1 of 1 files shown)


zachdaniel:

PR #312, opened 2025-05-07
fix: refactor internal `AshGraphql.Resource.mutation/6`

vonagam:
Just refactor, no fixes here. Mutation definitions were mostly the same and the same code was repeated multiple times. Difference between create/update/destroy was only absence of identity/read arguments for create.

But note for future - I noticed that `argument_input_types` does not apply to read action arguments right now.

### DIFF
--- a/lib/resource/resource.ex
+++ b/lib/resource/resource.ex
@@ -703,218 +703,51 @@ defmodule AshGraphql.Resource do
     end)
   end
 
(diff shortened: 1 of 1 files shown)


zachdaniel:

We should open an issue to track the `argument_input_types` not working for read actions 😄

vonagam:
Ok, if I will not fix it with my other changes by the end of the week - I'll open an issue.

PR #314, opened 2025-05-11
improvement: add `args` option to mutations

vonagam:
Adds `args` option to specify action attributes/arguments that should get their own arguments in the mutation method instead being passed in an input object.

(Also small fix - those [lines](https://github.com/ash-project/ash_graphql/blob/1e5149b0379d2815a4ced2f9f8871761a5b58832/lib/resource/resource.ex#L850-L855) that came from logic before my refactor meant that for generic action it's arguments were used to check if input needs to be non-null, but `action.arguments` and `mutation_fields` are different things - action arguments won't contain `Absinthe.Blueprint.Schema.FieldDefinition`, so based on that I assume that generic action input object was always nullable.)

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -6,6 +6,7 @@ spark_locals_without_parens = [
   actions: 1,
   actor: 1,
   allow_nil?: 1,
+  args: 1,
   argument_input_types: 1,
   argument_names: 1,
   as_mutation?: 1,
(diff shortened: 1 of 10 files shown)


zachdaniel:
This is great, something folks have asked for many times. There is currently a failing test in `main` which I'll address, but can you run `spark.formatter` to get the build passing?

zachdaniel:
looks like no tests failing after all. I think they only fail against Ash main 😓. That's fine though, still investigating that bug.

zachdaniel:

PR #317, opened 2025-05-16
Add meta to context

xantrac:
### Contributor checklist

In line with what's available in Absinth, this PR will allow the user to define a meta attribute on queries, mutations and subscriptions. The kwargs will be available as part of the as context (not Absinthe meta) during the resolution.

for example this can be used as follwing in a middleware

```
    queries do
      read_one :verify_public_event_token, :verify_token do
        meta(public: true)
      end
    end
```

and then in an Absinthe middleware 

```
  def call(resolution, _config) do
    meta = resolution.context.meta

    if meta[:public] do
      resolution
    else
      authenticate(resolution)
    end
  end
```


zachdaniel:
Is this PR ready for review? I see some IO.inspects etc.

xantrac:
> Is this PR ready for review? I see some IO.inspects etc.

this is r4r now

zachdaniel:
Please run `mix spark.cheat_sheets` and push the result of that up 🙇

xantrac:
> Please run `mix spark.cheat_sheets` and push the result of that up 🙇

Done, sorry I never worked with spark before

xantrac:
Let me know if is there anything else I need to get done on this

zachdaniel:

PR #319, opened 2025-05-22
fix: change default playground interface

Gazler:
Previously, the default playground interface was set to `:playground` which uses graphql-playground. However, this uses a deprecated JavaScript API which means that it does not work on modern versions of chrome. It also seems that the library is no longer maintained, so the chances of this being fixed in a future release are slim.

The result of this, is that tooltips do not disappear, making the application seem broken. Whilst this isn't a direct issue of this library, I think it gives a bad impression when things are broken out of the box.

This commit changes the generated interface to graphiql (using the `:simple` interface). There is also an `:advanced` interface option, but that also points to an unmaintained library which didn't work out of the box.

https://github.com/graphql/graphql-playground/issues/1429

Sorry @sevenseacat this might mean the screenshots in the book need updating.

### DIFF
--- a/lib/igniter.ex
+++ b/lib/igniter.ex
@@ -194,7 +194,7 @@ if Code.ensure_loaded?(Igniter) do
                   Absinthe.Plug.GraphiQL,
                   schema: Module.concat(["#{inspect(schema_name)}"]),
                   socket: Module.concat(["#{inspect(socket_name)}"]),
-                  interface: :playground
+                  interface: :simple
 
           forward "/",
             Absinthe.Plug,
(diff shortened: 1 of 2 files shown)


zachdaniel:
Then perhaps we should just poke them to update graphiql 😂

With that said, it looks like there are also other issues w/ `:simple`? https://github.com/absinthe-graphql/absinthe_plug/issues/303

Gazler:
I believe that the `graphiql-workspace` is used by `:advanced` (which is default) and hasn't been maintained for an even longer time!

https://github.com/OlegIlyenko/graphiql-workspace

zachdaniel:
🤔 AFAIK there are graphql playground tools that you can use and just point at an API, so maybe we should just encourage people to use those. Like this kind of thing: https://studio.apollographql.com/sandbox/explorer/?

So maybe we should generate the code, commented out, link to a description of why you *can* use these but you should prefer some online explorer? Not sure.

Gazler:
You can do something like that. It'll require CORS to be set up correctly to allow access though. I think you can embed apollo sandbox locally too, but that's no different to vendoring graphiql really.

PR #320, opened 2025-05-28
make pagination metadata more robust for offset based pagination

xantrac:
### Contributor checklist

### DIFF
--- a/lib/resource/resource.ex
+++ b/lib/resource/resource.ex
@@ -3832,6 +3832,46 @@ defmodule AshGraphql.Resource do
     ]
   end
 
+  defp add_pagination_metadata(fields, schema, true) do
+    [
+      %Absinthe.Blueprint.Schema.FieldDefinition{
+        description: "Whether or not there is a next page",
+        identifier: :has_next_page,
+        module: schema,
+        name: "has_next_page",
+        __reference__: ref(__ENV__),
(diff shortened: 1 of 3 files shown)


lib/resource/resource.ex
```diff
@@ -3865,19 +3905,10 @@ defmodule AshGraphql.Resource do
                 of_type: type
               }
             }
-          },
-          %Absinthe.Blueprint.Schema.FieldDefinition{
```
xantrac, line 3869: I removed this field definition since it doesn't see be used anywhere

zachdaniel: When you say its not used anywhere, wouldn't it have been used by the current type?

It remaps `more?` to `has_next_page`. Your current code handles that though so it should be fine to remove.

```elixir
     %Absinthe.Blueprint.Schema.FieldDefinition{


         description: "Whether or not there is a next page",
         identifier: :has_next_page,
         module: schema,
         name: "has_next_page",
         __reference__: ref(__ENV__),
         type: %Absinthe.Blueprint.TypeReference.NonNull{of_type: :boolean}
       },
```

zachdaniel (approved):

zachdaniel:
The compile error here is from a change made to support OTP 28, requires Ash to be upgraded. Does `mix check` pass for you locally? If so I'll merge and confirm

xantrac:
> mix check
is green for me
image

Issue #330, opened 2025-06-15
Support `sortable_fields` to override per-attribute `sortable?` flags


jechol:
**Is your feature request related to a problem? Please describe.**
Currently, GraphQL sorting behavior is determined solely based on the per-attribute `sortable?` flags.
While `filterable_fields` allows overriding the default `filterable?` behavior, there is no corresponding `sortable_fields` option to override `sortable?`.

**Describe the solution you'd like**
I would like to have a sortable_fields option, similar to filterable_fields, to explicitly control which fields are exposed as sortable in the GraphQL API—regardless of their underlying attribute settings.

**Describe alternatives you've considered**

**Additional context**

zachdaniel:
PRs welcome!

PR #331, opened 2025-06-17
Add sortable_fields option

jechol:
PR for https://github.com/ash-project/ash_graphql/issues/330

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/support/resources/tag.ex
+++ b/test/support/resources/tag.ex
@@ -10,6 +10,7 @@ defmodule AshGraphql.Test.Tag do
     type(:tag)
 
     filterable_fields [:name]
+    sortable_fields [:name]
 
     queries do
       get :get_tag, :read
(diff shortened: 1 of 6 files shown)


test/support/resources/tag.ex
```diff
@@ -10,6 +10,7 @@ defmodule AshGraphql.Test.Tag do
     type(:tag)
 
     filterable_fields [:name]
+    sortable_fields [:name]
```
zachdaniel, line 13: would be nice to add a test that shows not being able to sort on a different field.

jechol: I’d like to add a test, but I’m not sure which layer would be appropriate for it. I tried to find an example of testing the schema itself, but couldn’t locate one.

jechol: Thanks for pointing me to that PR — super helpful!
I’ve added a test that verifies filterable_fields and sortable_fields at the schema level.

PR #337, opened 2025-07-02
fix: fix :none relationship pagination strategy, and improve tests for it

jsw800:
# Contributor checklist

So it turns out that the test I wrote for this was broken, and also it turns out that the previous implementation was broken as well, since the relationship logic hooked into some "default" pagination logic that expects `nil` for no pagination. This fixes the `:none` strategy, and adds some better tests.

- [X] Chores
- [X] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel (approved):

zachdaniel:

PR #341, opened 2025-07-14
Add domain-level pubsub configuration for subscriptions

barnabasJ:
This allows users to set pubsub at the domain level instead of requiring it at every resource level. Resources can still override domain-level pubsub with their own configuration.


lib/resource/verifiers/verify_subscription_actions.ex
```diff
@@ -53,4 +58,26 @@ defmodule AshGraphql.Resource.Verifiers.VerifySubscriptionActions do
         path: [:graphql, :subscriptions, subscription.name, :read_action]
     end
   end
+
+  defp verify_pubsub(dsl, _subscriptions) do
+    resource_pubsub = AshGraphql.Resource.Info.subscription_pubsub(dsl)
+
+    domain_pubsub =
```
zachdaniel, line 65: We can't make a resource depend on the domain like this without risking compile time issues. This will need to be a verifier on the domain.

zachdaniel (changes requested):

zachdaniel:
@barnabasJ merge at your liesure.

PR #343, opened 2025-07-21
Add failing test for a calculation with a :struct return type

skanderm:
Types don't seem to get pulled in when using a `:struct` return type in a calculation with an `instance_of` constraint. I ran into this as a more complicated circumstance (a `calculation -> struct with instance_of constraint (another resource) -> calculation -> union NewType -> array -> struct`), but this might be the core of it!

```
# 1
%{
  errors: [
    %{message: "Field \"structCalc\" must not have a selection since type \"JsonString\" has no subfields.", locations: [%{line: 3, column: 5}]},
    %{message: "Cannot query field \"some\" on type \"JsonString\".", locations: [%{line: 4, column: 7}]}
  ]
}
```

### DIFF
--- a/lib/ash_graphql.ex
+++ b/lib/ash_graphql.ex
@@ -748,6 +748,16 @@ defmodule AshGraphql do
         type = Ash.Type.NewType.subtype_of(type)
         nested_attrs(type, all_domains, constraints, already_checked)
 
+      type == Ash.Type.Struct && Keyword.has_key?(constraints, :instance_of) ->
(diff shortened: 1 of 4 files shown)


zachdaniel:
Thanks for the test! I don't have time to investigate this right now. The issue will likely arise from somewhere in `AshGraphql.all_attributes_and_arguments` not traversing struct types.

skanderm:
Closing in favor of #444!

PR #366, opened 2025-09-26
Allow for typed struct constraints instead of using JsonString

DGollings:
AI generated, but manually tested using our code as input and we validated the schema.

Basically allows for 
```elixir
action :do_something do
      argument :some_struct, :struct,
        allow_nil?: false,
        constraints: [instance_of: SomeStruct]
end
```

which then results in the argument being set as
```graphql
  someStruct: SomeStructInput!
```

instead of 
```graphql
  someStruct: JsonString!
```

including a edge case where we had resources SomeStruct and SomeStructInput, where the former does not accidentally override the latter during generation
.
Also handles relationships of SomeStruct, so you see AnotherStruct within SomeStruct if relevant.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [X] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/resource/resource.ex
+++ b/lib/resource/resource.ex
@@ -2890,20 +2890,17 @@ defmodule AshGraphql.Resource do
     end
   end
 
-  # Generate input type definition for resources that can be used as struct types
-  # Only generates if no conflicting type exists
+  @doc """
+  Generate input type definition for resources that can be used as struct types.
+  Only generates if no conflicting type exists.
+  """
   def struct_input_type_definition(resource, _domain, all_domains, schema) do
     if AshGraphql.Resource.Info.type(resource) do
-      # Generate input type based on the resource's public attributes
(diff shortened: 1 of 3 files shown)


DGollings:
sure, but I am wondering if it could not be handled automatically, ie: improve the conflict checking code and other possible edge cases.
It already works for a project with 11 domains with 68 resources, all interlinked and two of which already conflict (and added a handler for that)

As in, we might be close already to a less manual solution

zachdaniel:
What I'm getting at is not necessarily about the reliability of conflict checking. Absinthe checks for conflicting types automatically. What I'm saying is that the input type for a resource should be user-determined, not derived from the read-representation type. And the best way for the user to decide type names for types is with `Ash.Type.NewType`.

DGollings:
sure, so what do you think of #367 instead?

DGollings:
with #368 as a potential minor DX upgrade

PR #367, opened 2025-09-27
added documentation for how to add and use structured GraphQL types for standalone struct arguments

DGollings:
# Contributor checklist

- [ ] Chores
- [X] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/documentation/topics/use-struct-types-with-graphql.md
+++ b/documentation/topics/use-struct-types-with-graphql.md
@@ -1,0 +1,111 @@
+# Use Struct Types with GraphQL
+
+When you need to pass structured data as standalone arguments (not creating relationships), custom struct types provide structured GraphQL input validation instead of generic JSON strings.
+
+## When You Need This
+
+**Relationship arguments** automatically get structured inputs:
+```elixir
+# ✅ Gets structured CreateAddressInput automatically
+argument :address, :map
+change manage_relationship(:address, type: :direct_control)
+```
+
+**Standalone data arguments** fall back to JsonString:
+```elixir
+# ❌ Becomes JsonString! (no structure validation)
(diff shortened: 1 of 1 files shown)


DGollings:
Sure, but whilst trying to implement it I have to say I prefer Ash.Type.NewType, as it creates a direct binding to the resource its referring to (we're not trying to create a new type here, but using an existing type elsewhere). 

With the current solution we don't get type conflicts (forcing a rename) and it automatically stays in sync.

Am I misunderstanding how `Ash.TypedStruct` works?

zachdaniel:
Oh, right my bad. Brain fart.

zachdaniel:
This looks good but I think we should lower the verbosity. like the table/trouble shooting section.

DGollings:
We could lower it, I'll have a look, but I'd have to add that both that table and the troubleshooting steps would have saved me several hours of searching :)

documentation/topics/use-struct-types-with-graphql.md
```diff
@@ -0,0 +1,107 @@
...
+   ```elixir
+   graphql do
+     mutations do
+       create :create_ticket_metadata, :create
+     end
+   end
+   ```
+3. Run `mix ash.codegen` to regenerate schema
+
+**Note:** Using `:struct` with `instance_of` directly also falls back to JsonString. Always wrap in a custom type for structured validation.
```
zachdaniel, line 107:
```suggestion
**Note:** Using `:struct` with `instance_of` directly also falls back to JsonString when used as an input. Always wrap in a custom type for structured validation.
```

zachdaniel:
Looks great! One last comment and lets ship it 🔥

PR #380, opened 2025-10-30
Fix aggregate queries failing when filter references parent fields

### LINKED ISSUE #379: GraphQL list queries fail to load aggregates with parent filter references
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

- Elixir: 1.18.3
- Ash: 3.7.6
- AshGraphql: 1.8.3
- AshPostgres: 2.6.23
- AshSql: 0.3.9

### Operating system

macOS 15.6.1

### Current Behavior

GraphQL list queries fail when loading aggregates that reference parent fields in their filter expressions.

Error: `ERROR 42703 (undefined_column) column s0.last_read_msg_id does not exist`

The generated SQL's subquery `s0` doesn't include the field referenced in the aggregate filter.

Important: This ONLY affects aggregates in `list` queries. Everything else works:
- Resource-level queries (read/read_one) ✅
- GraphQL `get` queries ✅
- has_many relationships with identical filters ✅

### Reproduction

Full reproduction repository: https://github.com/jechol/ash-graphql-aggregate-bug

Expected: 1 failure out of 8 tests (GraphQL list query with aggregate)

The repository includes:
- Minimal schema (Chat, Message)
- Aggregate using parent(last_read_msg_id) in filter
- has_many with identical filter (works fine)
- Tests comparing all scenarios

### Expected Behavior

GraphQL list queries should successfully load aggregates that reference parent fields.

jechol:
## Fixes #379

GraphQL list queries were failing with 'column does not exist' errors when loading aggregates that use parent() references in their filters.

The issue occurred because the parent fields referenced in aggregate filters were not being selected in the main query's subquery, but were being referenced in the LATERAL join for the aggregate calculation.

For example, an aggregate filter like 'expr(parent(last_read_msg_id))' would generate SQL that references s0.last_read_msg_id in the LATERAL join, but the field wasn't selected in the s0 subquery.

Changes:
- Extract parent-referenced fields from aggregate filters and load them alongside the aggregate
- Add extract_parent_fields/1 to traverse filter AST and identify parent() expressions  
- Add extract_refs_from_parent/1 helper to extract field names only from within parent() contexts

This ensures parent-referenced fields are included in the query selection, making them available for aggregate calculations in LATERAL joins.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/graphql/resolver.ex
+++ b/lib/graphql/resolver.ex
@@ -2185,7 +2185,9 @@ defmodule AshGraphql.Graphql.Resolver do
     Enum.flat_map(fields, fn selection ->
       cond do
         aggregate = Ash.Resource.Info.aggregate(resource, selection.schema_node.identifier) ->
-          [aggregate.name]
+          # Load the aggregate and any parent-referenced fields
+          parent_fields = extract_parent_fields(aggregate.filter)
+          [aggregate.name | parent_fields]
 
         calculation = Ash.Resource.Info.calculation(resource, selection.schema_node.identifier) ->
           arguments =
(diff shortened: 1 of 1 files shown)


zachdaniel:
I'm 99% sure that the fix to this issue would be in `ash_sql`, not here. For callers, it should never matter if you do or don't select fields that aggregates depend on. Can we instead create a failing test in `ash_postgres` and/or a fix in `ash`?

jechol:
I actually had the same feeling that this should be fixed in ash_sql/ash_postgres. I tried to fix it there first, but it was too complex for me to figure out properly. 

For now, I'll use this as a workaround in my project.

PR #386, opened 2025-11-14
clear out managed relationship requirements process variable once it's been used

jsw800:
My team has been having some schema instability in certain cases in our codebase. I did some debugging today and it seemed like some managed relationship inputs were somehow getting associated to the wrong domain in certain cases. I'm fairly certain this comes down to the fact that this process variable (which looks very hacky to me 😅 ) is only ever being added to, but never cleared out once it's been processed.

I'm not 100% sure if this fix is correct so I'd appreciate if someone with better familiarity with this code gave this a close review.

# Contributor checklist

  - [ ] I honestly don't know how to put together a replication of this bug, sorry
- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash_graphql.ex
+++ b/lib/ash_graphql.ex
@@ -260,9 +260,14 @@ defmodule AshGraphql do
                 Absinthe.Blueprint.add_field(blueprint, "RootSubscriptionType", subscription)
               end)
 
+            managed_relationship_requirements =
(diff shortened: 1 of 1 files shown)


zachdaniel:
Naturally a better fix would be to thread these through, not do it state fully. With that said, I'm glad you found this tricky bug 🙇 

PR #406, opened 2026-02-23
fix: don't pass select to mutations, fixes notification data

DGollings:
## Problem

AshGraphQL passes select: get_select to create/update/destroy mutations, restricting the Ash operation to only the fields the GraphQL client requested. Since Ash v3.4.8's action_select system, this masks all
unrequested attributes with %Ash.NotLoaded{} — including on notification data sent to Ash notifiers.

This means any mutation like:

```elixir
mutation { deliverParcel(id: "...") { result { id status } } }
```

produces notifications where order_id, company_id, and other foreign keys are Ash.NotLoaded, crashing downstream PubSub subscribers and GenServers that rely on those fields.

The issue became more visible after ash-project/ash#2544 (manage_relationship returning {:not_atomic}), which forces more actions through the non-atomic bulk path where the masking is applied more aggressively.

## Fix

Remove the select option from all three mutation paths (create, update, destroy). Ash defaults to loading all select_by_default attributes, which is what notifiers need. Absinthe already filters the GraphQL response to only
  the requested fields.

## Test plan

- Added NotificationSpy test notifier that captures raw Ash notifications
- Tests verify that belongs_to foreign keys are loaded in notification data even when the GraphQL query only requests id
- Verified the update test fails without the fix (actor_id becomes Ash.NotLoaded)
- Full test suite passes (293 tests, 0 failures)

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/graphql/errors.ex
+++ b/lib/graphql/errors.ex
@@ -111,8 +111,7 @@ defmodule AshGraphql.Errors do
     {mapped_path, _context} =
       Enum.reduce(all_segments, {[], initial_context}, fn segment, {acc, ctx} ->
         {resolved, next_ctx} = resolve_segment_with_context(segment, ctx)
-        acc = if resolved, do: acc ++ [resolved], else: acc
-        {acc, next_ctx}
+        {acc ++ [resolved], next_ctx}
       end)
 
     # Append field name from error map (resolved with current context).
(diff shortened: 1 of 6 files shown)


zachdaniel:
@DGollings I think this bug may have been fixed in Ash core recently. We should still pass `select` into these actions, if its breaking your notifications then the bug is elsewhere 😄

DGollings:
completely forgot this PR was even open :sweat_smile: , glad to hear it

PR #424, opened 2026-03-27
fix: scope AshTypes module to schema for multi-schema support

### LINKED ISSUE #422: AshTypes module conflict when same domain appears in multiple Absinthe schemas
## Context

I'm building a multi-schema GraphQL architecture where a single Phoenix app exposes **multiple Absinthe schemas** (subdomain schemas), each serving a different GraphQL endpoint. Each schema registers a subset of Ash domains via `use AshGraphql, domains: [...]`.

This is useful for large enterprise apps with 60+ domains — instead of one monolithic schema with thousands of types, we split into ~20 focused schemas (e.g. `finance`, `sales`, `travel`, `portal`).

**The problem arises when domains are shared across schemas.** For example, `Sales` domain is needed by both the `sales` schema and the `purchasing` schema (because Purchasing resources have `belongs_to` relationships pointing to Sales entities). Without including `Sales` in both schemas, Absinthe reports "type is not defined in your schema" errors.

## Reproduction

```elixir
# Schema A
defmodule MyApp.Schema.Sales do
  use Absinthe.Schema
  use AshGraphql, domains: [MyApp.Sales, MyApp.Pricing]
  # ...
end

# Schema B — also needs Pricing for cross-domain type references
defmodule MyApp.Schema.Purchasing do
  use Absinthe.Schema
  use AshGraphql, domains: [MyApp.Purchasing, MyApp.Pricing]
  # ...
end
```

## Issue 1: `{Domain}.AshTypes` module defined twice

`use AshGraphql` creates `Module.concat(domain, AshTypes)` for each domain ([ash_graphql.ex#L187](https://github.com/ash-project/ash_graphql/blob/main/lib/ash_graphql.ex#L187)). When `Pricing` appears in both schemas, `MyApp.Pricing.AshTypes` gets defined twice → compilation error:

```
error: cannot define module MyApp.Pricing.AshTypes because it is currently being defined in lib/schemas/sales.ex
```

**Suggested fix:** Guard with `unless Code.ensure_loaded?(Module.concat(domain, AshTypes))` before `defmodule`.

## Issue 2: Global types not deduplicated against existing types

When a schema uses `import_types` to bring in shared types (e.g. `mutation_error`, `sort_order`, `page_info`), and AshGraphql's `first?` domain also injects these via `global_type_definitions/2`, the types appear twice in the blueprint → Absinthe's `TypeNamesAreUnique` phase reports:

```
Absinthe type identifier :mutation_error is not unique.
```

The root cause is in [ash_graphql.ex#L356-L364](https://github.com/ash-project/ash_graphql/blob/main/lib/ash_graphql.ex#L356-L364) — `type_definitions ++ managed_relationship_types` are appended unconditionally without checking for existing types in `schema_def.type_definitions`.

**Suggested fix:** Deduplicate before appending:

```elixir
new_defs =
  List.update_at(blueprint.schema_definitions, 0, fn schema_def ->
    existing_ids = MapSet.new(schema_def.type_definitions, & &1.identifier)
    new_types =
      (type_definitions ++ managed_relationship_types)
      |> Enum.reject(fn type -> MapSet.member?(existing_ids, type.identifier) end)
    %{schema_def | type_definitions: schema_def.type_definitions ++ new_types}
  end)
```

## Related

- #163 — "Type deduplication quirks" (closed, same root cause family)

## Workaround

We currently maintain a [fork](https://github.com/biantaishabi2/ash_graphql/tree/unibo-patches) with both fixes applied on top of v1.8.5. Happy to open a PR if the approach looks reasonable.

biantaishabi2:
## Summary

When the same Ash domain is registered in multiple Absinthe schemas (e.g., splitting a large app into Sales/Purchasing/Finance endpoints with shared domains for cross-domain `belongs_to` references), `Module.concat(domain, AshTypes)` generates the same module name regardless of which schema is compiling, causing module redefinition conflicts.

This changes the module name to `Module.concat([__MODULE__, domain, AshTypes])` as suggested in #422, scoping each generated module to its parent schema.

This also fixes the `first?` / builtin types issue (#422 issue 2) as a side effect: the root cause was that the module conflict caused the second schema to skip the entire `for` loop body, preventing `first?` from working. With per-schema modules, each schema's `for` loop executes fully, and `first?` (already per-schema via `List.update_at(0, ...)`) works correctly.

## Changes

- `defmodule Module.concat(domain, AshTypes)` → `defmodule Module.concat([__MODULE__, domain, AshTypes])`
- `@pipeline_modifier Module.concat(domain, AshTypes)` → `@pipeline_modifier Module.concat([__MODULE__, domain, AshTypes])`

## Test plan

- All 359 existing tests pass
- Verified with a production app using ~20 subdomain schemas with shared domains — compiles without conflicts

Closes #422

### DIFF
--- a/lib/ash_graphql.ex
+++ b/lib/ash_graphql.ex
@@ -199,7 +199,7 @@ defmodule AshGraphql do
       end
 
       for {domain, resources, first?} <- domains do
-        defmodule Module.concat(domain, AshTypes) do
+        defmodule Module.concat([__MODULE__, domain, AshTypes]) do
           @moduledoc false
           alias Absinthe.{Blueprint, Phase, Pipeline}
 
@@ -386,7 +386,7 @@ defmodule AshGraphql do
           Code.eval_quoted(auto_import_types, [], __ENV__)
         end
 
-        @pipeline_modifier Module.concat(domain, AshTypes)
+        @pipeline_modifier Module.concat([__MODULE__, domain, AshTypes])
       end
     end
   end
--- a/test/multi_schema_test.exs
+++ b/test/multi_schema_test.exs
@@ -1,0 +1,42 @@
+# SPDX-FileCopyrightText: 2026 ash_graphql contributors
+#
+# SPDX-License-Identifier: MIT
+
+defmodule AshGraphql.MultiSchemaTest do
+  @moduledoc """
+  Tests that the same domain can be registered in multiple Absinthe schemas
+  without module conflicts or missing builtin types.
+  """
+  use ExUnit.Case, async: false
+
+  # Two schemas sharing AshGraphql.Test.Domain — this should compile
+  # without "cannot define module" errors or missing type errors.
+
+  test "same domain in multiple schemas compiles without conflict" do
+    # If we got here, both schemas compiled successfully.
+    # AshGraphql.Test.Schema and AshGraphql.Test.MultiSchema.SchemaB
+    # both include AshGraphql.Test.Domain.
+    assert Code.ensure_loaded?(AshGraphql.Test.Schema)
+    assert Code.ensure_loaded?(AshGraphql.Test.MultiSchema.SchemaB)
+  end
+
+  test "each schema has its own AshTypes module" do
+    # Schema A's AshTypes (OtherDomain is in the primary schema's domain list)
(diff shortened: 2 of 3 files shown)


zachdaniel:

PR #425, opened 2026-03-27
fix: per-schema AshTypes + structural type dedup for multi-schema support

biantaishabi2:
## Summary

This is an alternative to #424 that includes both fixes for #422:

1. **Per-schema AshTypes module** (same as #424): `Module.concat([__MODULE__, domain, AshTypes])` scopes each generated module to its parent schema, preventing module redefinition conflicts when the same domain appears in multiple schemas.

2. **Structural type deduplication**: When appending `type_definitions` to the blueprint, reject types only when an existing type has the same identifier AND is structurally equal (`existing == type`). Types with the same name but different structure are left for Absinthe's `TypeNamesAreUnique` phase to diagnose.

This addresses both points from #422. The dedup approach follows your suggestion of checking "100% exactly equal" rather than just identifier matching.

## Changes

- `Module.concat(domain, AshTypes)` → `Module.concat([__MODULE__, domain, AshTypes])` (2 locations)
- `type_definitions` append: reject structurally identical duplicates before appending

## Test plan

- All 359 existing tests pass
- Verified with a production app using ~20 subdomain schemas with shared domains

See #424 for the minimal version (without dedup) if you prefer a smaller change.

Related: #422

### DIFF
--- a/lib/ash_graphql.ex
+++ b/lib/ash_graphql.ex
@@ -199,7 +199,7 @@ defmodule AshGraphql do
       end
 
       for {domain, resources, first?} <- domains do
-        defmodule Module.concat(domain, AshTypes) do
+        defmodule Module.concat([__MODULE__, domain, AshTypes]) do
           @moduledoc false
           alias Absinthe.{Blueprint, Phase, Pipeline}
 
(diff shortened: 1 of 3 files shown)


biantaishabi2:
Closing in favor of a separate independent PR — splitting issue 1 and issue 2 fixes into independent PRs so they can be merged independently.

PR #431, opened 2026-05-21
fix: use pattern matching in resolve_array_element_segment

DGollings:
Obviously generated PR message, but I've been running these fixes for weeks/months in prod

## Summary

`AshGraphql.Errors.resolve_array_element_segment/2` calls `elem(context.type, 2)` on what is a 2-tuple `{:array, inner_type}`. Index `2` is out of range for a 2-tuple, so `elem/2` raises `ArgumentError` whenever the error formatter walks a path that crosses an array argument (e.g. `[:comments, 0, :text]`).

The result in production is that the GraphQL error formatter itself crashes mid-formatting, and the underlying business error never reaches the client — the user just sees a 500.

This PR replaces the `elem/2` call with a pattern match that extracts the inner type safely, and adds a regression test that exercises the path through a `{:array, :map}` argument.

### DIFF
--- a/lib/graphql/errors.ex
+++ b/lib/graphql/errors.ex
@@ -253,7 +253,11 @@ defmodule AshGraphql.Errors do
 
   defp resolve_array_element_segment(segment, context) do
     # Context type is {:array, elem_type}; after index we're "in" element.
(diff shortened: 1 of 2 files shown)


zachdaniel (approved):

zachdaniel:

PR #432, opened 2026-05-21
fix: don't pass select to mutations, fixes notification data

DGollings:
Obviously generated PR message, but I've been running these fixes for weeks/months in prod

## Summary

AshGraphQL currently passes `select: get_select(...)` to create/update/destroy mutations, derived from the fields requested in the GraphQL query. Ash's `action_select` system then masks all unrequested attributes with `%Ash.NotLoaded{}` on the result — **including the result handed to notifiers**.

Downstream consumers (PubSub subscribers, custom `Ash.Notifier` modules, GenServers) that read attributes such as foreign keys from `belongs_to` relationships then crash, because those fields are `%Ash.NotLoaded{}`. From the caller's perspective the mutation succeeded, but every notifier downstream blows up.

This PR removes the `select` option from the create, update, and destroy mutation paths in `lib/graphql/resolver.ex`. Ash defaults to loading all `select_by_default?: true` attributes, which is what notifiers expect. Absinthe still filters the GraphQL response to only the fields the client requested, so this is transparent to clients.

## Tests

- New `test/mutation_notification_test.exs` verifies that a notifier receives a fully-loaded record after each of `create`, `update`, and `destroy` mutations, even when the GraphQL query only selects a single field.
- New `test/support/notification_spy.ex` is a minimal notifier used by that test.

## Trade-offs / discussion

This changes the default attribute load behaviour for mutations. Specifically: mutations will now load all `select_by_default?: true` attributes on the result row regardless of what the client requested. For most resources this is the same set that's already loaded by reads, so the practical cost is small, but it is technically a behaviour change.

An alternative would be to keep `select` for the response and separately load all attributes before notifiers run — happy to take that direction instead if maintainers prefer to preserve the current default.

### DIFF
--- a/lib/graphql/resolver.ex
+++ b/lib/graphql/resolver.ex
@@ -1560,7 +1560,6 @@ defmodule AshGraphql.Graphql.Resolver do
               actor: Map.get(context, :actor),
               authorize?: AshGraphql.Domain.Info.authorize?(domain)
             )
-            |> select_fields(resource, resolution, type_name, ["result"])
             |> load_fields(
               [
                 domain: domain,
@@ -1693,10 +1692,6 @@ defmodule AshGraphql.Graphql.Resolver do
                   read_action: read_action,
                   domain: domain,
                   actor: Map.get(context, :actor),
-                  select:
-                    get_select(resource, resolution, mutation_result_type(mutation_name), [
-                      "result"
-                    ]),
                   load:
                     get_loads(
                       [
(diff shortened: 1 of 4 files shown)


zachdaniel:
Passing `select` to a mutation should not affect notifications. I believe we fixed this in a recent update to `ash`. If not, we need to fix it there. Can you take a look there please? I'm going to close this for now but we can reopen if that proves not to be the case.

PR #434, opened 2026-05-27
Implement :auto for calculations

olivermt:
I have dogfooded this a bit and it seems to work as-is, removing the need for quite a bit of typing of projection style calculations.


olivermt:
This was included in the protected field PR

PR #435, opened 2026-05-28
Feature: Add protected fields as first class citizen

olivermt:
## Summary

Adds `field_policy_mode` for GraphQL resources so fields that may be hidden by authorization can be represented more intentionally in the generated schema.

## Changes

- Add `field_policy_mode :nullable` to automatically make protected fields nullable.
- Add `field_policy_mode :materialized` to expose protected fields as value-or-`ForbiddenField` unions.
- Materialize singular relationships with `allow_forbidden_field? true` as destination-or-`ForbiddenField` unions.
- Use Ash’s authorizer protected-field metadata for detection.
- Warn when custom authorizers are used with `field_policy_mode`, since they must report protected fields for schema generation to reflect them.
- Document nullable/materialized modes and the singular relationship limitation.
- Update Ash lockfile to Hex `ash 3.27.6`.

## Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [x] Update dependencies


zachdaniel:
I might call this something like `forbidden_field_mode` or something like that? Because more than field policies can produce this scenario, including custom authorizers.

lib/graphql/resolver.ex
```diff
@@ -3041,12 +3054,46 @@ defmodule AshGraphql.Graphql.Resolver do
...
+    }
+  end
+
+  defp materialize_forbidden_field(%Ash.ForbiddenField{} = forbidden, resolution) do
+    field = resolution.definition.name || forbidden.field
+
+    %{
+      __field_policy_type__: :forbidden_field,
+      field: if(field, do: to_string(field)),
+      type: if(forbidden.type, do: to_string(forbidden.type)),
```
zachdaniel: We don't expose the internal type of fields, i.e aggregate/calculations etc. That is an internal implementation detail.

lib/resource/verifiers/verify_field_dependencies.ex
```diff
@@ -272,6 +273,33 @@ defmodule AshGraphql.Resource.Verifiers.VerifyFieldDependencies do
...
+
+        if Enum.empty?(custom_authorizers) do
+          warnings
+        else
+          [custom_authorizers_warning(resource, mode, custom_authorizers) | warnings]
+        end
+    end
+  end
+
+  defp custom_authorizers_warning(resource, mode, custom_authorizers) do
```
zachdaniel, line 296: Unless I'm missing something, this warning is shown unconditionally, but it should exclude the authorizers that correctly implement the required callback?

zachdaniel (changes requested):
Have a few comments, in general this looks good.

lib/graphql/resolver.ex
olivermt: ack

lib/resource/verifiers/verify_field_dependencies.ex
olivermt: ack

olivermt:
> I might call this something like `forbidden_field_mode` or something like that? Because more than field policies can produce this scenario, including custom authorizers.

done now, ready for review agai

zachdaniel (dismissed):

zachdaniel:
This also appears to have merge conflicts.

olivermt:
Checks confirmed locally

zachdaniel (approved):

Issue #354, opened 2025-08-12
Subscription DSL filter causes filtered events to become errors


justin-m-morgan:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

Elixir - 1.18.4
Erlang - 28.0.2
Ash - 3.5.34
AshGraphql - 1.8.2
Absinthe - 1.7.10

### Operating system

MacOS 15.6

### Current Behavior

When using the subscription DSL filter, when a filter is applied, the events that out to be excluded are still sent with a `nil` payload and an error `code: not_found`.

Example: 

```gql
    subscription($filter: TestResourceFilterInput) {
      testResourceCreated(filter:$filter) {
        created {
          id
          choice
        }
      }
    }
```

```elixir 
%Phoenix.Socket.Message{
topic: "__absinthe__:doc:unRcSrcuHkiBCReefuo40W42GErDK62KmQKrtqYnwyU=:1E9C1D467CBCCEB83F768CA9C9AD88068931E40CF6C347A4D1FB07F29F3762C0", 
event: "subscription:data", 
payload: %{result: %{data: %{"testResourceCreated" => nil}, 
errors: [%{code: "not_found", message: "could not be found", path: ["testResourceCreated"], fields: [], vars: [], short_message: "could not be found", locations: [%{line: 2, column: 3}]}]}, 
subscriptionId: "__absinthe__:doc:unRcSrcuHkiBCReefuo40W42GErDK62KmQKrtqYnwyU=:1E9C1D467CBCCEB83F768CA9C9AD88068931E40CF6C347A4D1FB07F29F3762C0"}, ref: nil, join_ref: nil
} 
```

In the gql playground, this results in just a `null` event without the included error message which is particularly unhelpful.

```json
{
  "data": {
    "testResourceCreated": null
  }
}
```

### Reproduction

[Reproduction Repo](https://github.com/justin-m-morgan/ash_gql_sub_bug)

### Expected Behavior

When a filter is applied to a subscription, excluded events should not be sent to the subscriber as error events with no data payload.

justin-m-morgan:
@barnabasJ pointed out in the ash discord that a likely starting point of investigation is the `should_send?` function in the batcher: https://github.com/ash-project/ash_graphql/blob/d724c46d7d44cd516e26155d144ee7fe3f569867/lib/subscription/batcher.ex

[...]

PR #442, opened 2026-06-04
fix subscriptuon bug

aheiner2001:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/igniter.ex
+++ b/lib/igniter.ex
@@ -267,7 +267,7 @@ if Code.ensure_loaded?(Igniter) do
           igniter
           |> Igniter.Project.Module.find_and_update_module!(endpoint, fn zipper ->
             with {:ok, zipper} <- Igniter.Code.Module.move_to_use(zipper, Phoenix.Endpoint) do
-              {:ok, Igniter.Code.Common.add_code(zipper, "use Absinthe.Phoenix.Endpoint")}
+              {:ok, Igniter.Code.Common.add_code(zipper, "use AshGraphql.Subscription.Endpoint")}
             end
           end)
           |> Igniter.Project.Module.find_and_update_module!(endpoint, fn zipper ->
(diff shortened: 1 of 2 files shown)


zachdaniel:
@barnabasJ could you potentially take a look at this? Looks reasonable to me.

barnabasJ (commented):
@aheiner2001 the new test is already green with the resover / batcher implementation from main. could you describe what problem you were trying to solve exactly?

aheiner2001:
@barnabasJ .I was fixing issue #354: when a subscription DSL filter excludes an event, the subscriber should get nothing, not data: null with a not_found error. On main, the new test already passes because the batcher suppresses not_found before publish, even though the resolver still treats filter exclusion as NotFound. My change makes exclusion mean “drop” in the resolver ({:ok, nil}) and updates the batcher so null-only payloads aren’t sent either — those two parts need to work together.
I used the reproduction repo provided

barnabasJ:
> NotFound

Ah, I missed the referenced issue, I'll have another look over the weekend

barnabasJ:
@aheiner2001 I had another look at the issue as well, but couldn't really reproduce it. Were you able to make it fail locally the way the issue is described?

I don't necessarily see a problem with the changes, but it would be great to have a test that actually triggers the issue mentioned.

aheiner2001:
@barnabasJ 
Yes, I was able to reproduce the error using the repoduction repository. Here is the screen shot of the tests. 

From what I can see, this is the initial bug reported. 
image

barnabasJ:
@aheiner2001 looked at the reproduction repo, and I think the actual issue is in the igniter installer
https://github.com/ash-project/ash_graphql/blob/a0d8b9d1f49c8122ac4dea573e7f0ed911aeab4c/lib/igniter.ex#L270-L270 the installer adds the regular Absinthe endpoint, but it should add `AshGraphql.Subscription.Endpoint`, otherwise our filter logic will not run.

aheiner2001:
@barnabasJ  You're right — I've reverted my initial changes and implemented the ones you suggested.

barnabasJ (approved):

PR #446, opened 2026-06-12
Support field?: false calculations in GraphQL schema

jaylynntautu:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [X ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

## Summary

This updates GraphQL response schema generation to respect Ash calculations using `field?: false`.

Calculations marked with `field?: false` are now excluded from normal GraphQL response fields while still remaining available in filter and sort schema generation.

## Changes

* Updated `calculations/3` in `lib/resource/resource.ex` to exclude calculations where `field?` is false
* Replaced the old read test that allowed querying `nonFieldCalc`
* Added regression tests confirming:

  * `field?: false` calculations are not exposed on the `Post` response type
  * `field?: false` calculations are still included in `PostFilterInput`
  * `field?: false` calculations are still included in `PostSortField`

## Verification

Ran locally:

* `mix test`
* `mix format --check-formatted`
* `mix credo --strict`

Full test suite passing locally:

* 370 tests
* 0 failures

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -44,6 +44,7 @@ spark_locals_without_parens = [
   hide_inputs: 1,
   identity: 1,
   ignore?: 1,
+  include_non_field_calculations: 1,
   keyset_field: 1,
   list: 2,
   list: 3,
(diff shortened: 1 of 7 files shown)


zachdaniel:
So we actually *can* support these as fields in GraphQL, they just have to be picked from `record.calculations[name]` if loaded.

zachdaniel:
Maybe we should make that opt-in somehow though 🤔

jaylynntautu:
Ohhh okay I followed the original issue description and excluded them from the response schema while keeping filter/sort support. Since they can still be resolved through `record.calculations[name]`, would you like me to update this PR in that direction instead?

zachdaniel:
Yes please 🙇

zachdaniel:
Although, I think what we should do is hide them from the GraphQL by default and add a new DSL option to include them. Like `include_non_field_calculations: [:foo, :bar]`

PR #449, opened 2026-06-30
fix: preserve remaining GraphQL filters

ahey:
[This recent commit on ash_graphql main branch](https://github.com/ash-project/ash_graphql/commit/0a52795fd7d5896bc313a8c9ea99dc5260a6a3d5) has introduced a regression that was surfaced in my own project's test suite.

The previous empty-map match treated every remaining filter map as empty, so non-handler field filters were dropped before reaching Ash.

Cover handler, query, and subscription paths to prevent filtered GraphQL operations from returning records that should be excluded.

I used AI to diagnose and fix this issue, however I have personally checked and understand the changes.

# Contributor checklist

- [x] Chores

### DIFF
--- a/lib/graphql/filter_handlers.ex
+++ b/lib/graphql/filter_handlers.ex
@@ -25,7 +25,7 @@ defmodule AshGraphql.Graphql.FilterHandlers do
 
     filter_input =
       case remaining do
-        %{} -> nil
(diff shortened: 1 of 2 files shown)


zachdaniel (approved):

zachdaniel:

PR #456, opened 2026-07-14
Add custom mutation result names

### LINKED ISSUE #57: Allow customization of key `result(s)` with mutations
Right now when you create a mutation with AshGraphQL, the output data will be inside the `result` or `results` field.

Some people prefer to have the resource name there instead of a generic one.

For example, if the resource being returned is `Offer`, then the key would be `offer` or `offers`.

I would like to have some options that I add to the mutation to customize that key name.

I can achieve that if I create the whole mutation manually, but that kinda defeats the purpose of using Ash

jaylynntautu:
# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Summary

This PR adds support for customizing the GraphQL result field name returned by mutations.

Previously, all mutation payloads exposed the returned resource under the generic `result` field. With this change, users can optionally configure a custom result field name (for example, `offer` or `offers`) while preserving the existing `result` field as the default for backwards compatibility.

This change adds the new mutation DSL option, updates GraphQL schema generation to honor the configured field name, and includes tests covering both the default behavior and customized result names.

Closes #57

### DIFF
--- a/lib/graphql/resolver.ex
+++ b/lib/graphql/resolver.ex
@@ -1524,7 +1524,7 @@ defmodule AshGraphql.Graphql.Resolver do
            upsert_identity: upsert_identity,
            args: args,
            modify_resolution: modify
-         }, _relay_ids?}
+         } = mutation, _relay_ids?}
       ) do
     case handle_mutation_arguments(resource, action, nil, arguments, args) do
       {:ok, input} ->
@@ -1561,6 +1561,7 @@ defmodule AshGraphql.Graphql.Resolver do
             end
 
           type_name = mutation_result_type(mutation_name)
+          result_field_name = mutation_result_field_name(mutation)
 
           changeset =
             resource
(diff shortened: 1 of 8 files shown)


lib/graphql/resolver.ex
```diff
@@ -2829,6 +2829,11 @@ defmodule AshGraphql.Graphql.Resolver do
     |> elem(0)
   end
 
+  defp field_matches_path?(field, name) do
+    field.name == name or
+      (name == "result" and field.schema_node.identifier == :result)
```
zachdaniel: Something about this seems off. Been a while since I looked at this code unfortunately 😢 Could this somehow affect cases where there is an actual field somewhere called `:result`? I'm wondering if the right answer is actually to rename the result field's identifier to `:result` and handle the effects of that elsewhere. 

Granted this is less complex and requires less changes if we just change the presented name.

Could you think through the first bit above and LMK if there are cases where this is an issue?

jaylynntautu: Ok yes the identifier fallback could be too broad in cases where a resource has an actual field named :result.

I removed that fallback entirely, so field_matches_path?/2 now only matches field.name == name.

Instead of changing the schema identifier, I now thread the configured mutation result field name through the create/update/destroy selection and load traversal. The internal identifier remains :result, and resolver return maps remain %{result: value}.

I also added a regression test with a public resource attribute named :result to cover the collision case you called out.

zachdaniel:

PR #460, opened 2026-08-03
fix: don't raise building error paths for maps without field constraints

puyo:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Summary

Using `Ash.Changeset.add_error/3` or `Ash.Error.set_path/2` with a path that delves into `:map` types can swallow validation errors, returning a generic "Something went wrong" instead of the real error. This happens because `constraints` is a keyword list like `[preserve_nil_values?: false]`, and the `:fields` key is missing, so the `constraints["fields"]` fallback runs and raises on a keyword list looked up with a string key.

```elixir
iex(1)> x = [y: 1]
[y: 1]
iex(2)> x["hi"]
** (ArgumentError) the Access module supports only keyword lists (with atom keys), got: "hi"

If you want to search lists of tuples, use List.keyfind/3
    (elixir 1.20.2) lib/access.ex:362: Access.get/3
    iex:2: (file)
```

The same pattern appears four times (`:types`, `:fields` twice, `:items`). Only the `composite_kind/2` one raises today. The others have their atom key present, but I've changed all four since we never want to do `keyword_list["key"]`.

`add_error` and `set_path` do not prevent this kind of thing with their advertised type spec so this is something we walked into. This repo raises inside `build_error_path/5`, before both the resource and domain `error_handler` hooks, so there's no way to handle it from application code. #431 fixes a similar issue in this project, so I believe this belongs here.

The fix essentially removes the string fallback, which on a keyword list can only ever raise.

### DIFF
--- a/lib/graphql/errors.ex
+++ b/lib/graphql/errors.ex
@@ -185,7 +185,7 @@ defmodule AshGraphql.Errors do
   end
 
   defp resolve_union_segment(segment, context) do
-    types = context.constraints[:types] || context.constraints["types"] || %{}
+    types = Keyword.get(context.constraints, :types, %{})
     segment_atom = segment_to_atom(segment)
 
     config =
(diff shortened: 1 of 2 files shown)


zachdaniel (approved):

zachdaniel:

PR #461, opened 2026-08-12
fix: support calculations in identities used by graphql queries

maennchen:
A `get` query with `identity:` (and `update`/`destroy` mutations with an identity) looked up each identity key with
`Ash.Resource.Info.attribute/2`, which returns `nil` for a calculation. Building the argument then raised `KeyError: key :type not found in: nil` at schema compile time, so identities backed by a filterable calculation could not be used at all.

Use `Ash.Resource.Info.field/2` instead, which resolves attributes, calculations and aggregates alike and exposes the same `type` and `description`. The primary-key path is left alone, as it is always attributes.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/resource/resource.ex
+++ b/lib/resource/resource.ex
@@ -1396,15 +1396,15 @@ defmodule AshGraphql.Resource do
     |> Enum.find(&(&1.name == identity))
     |> Map.get(:keys)
     |> Enum.map(fn key ->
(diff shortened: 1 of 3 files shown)


zachdaniel (approved):

zachdaniel:

PR #464, opened 2026-08-29
feat: allow filter_handlers for public calculations

nallwhy:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

`filter_handlers` currently only accepts resource attributes. This PR allows handlers on public calculations as well, with the handler taking precedence over the calculation's normal filtering.

The motivating case is an encrypted field exposed as a calculation, where filtering happens through a separate blind-index attribute — e.g. AshCloak replaces the plaintext attribute with a non-filterable decryption calculation, while exact matching should query a private hash attribute:

```elixir
graphql do
  filterable_fields card_number: [:eq]

  filter_handlers card_number: [
    type: :string,
    handler: {MyApp.FilterHandlers, :blind_index, [:card_number_hash]}
  ]
end
```

```graphql
filter: { cardNumber: { eq: "1234567890" } }
```

This lets an application expose a logical field in GraphQL while keeping its underlying searchable representation private.


zachdaniel (approved):
LGTM, although we should do this same thing for public aggregates 😄

zachdaniel:

Issue #474, opened 2026-09-16
first: 0 and last: 0 are an unrendered error


grempe:
Versions: ash_graphql 1.11.0, absinthe 1.12.0, absinthe_plug 1.5.10, ash 3.33.3, ash_postgres 2.13.1, Elixir 1.20.1 on OTP 29. Present on the default branch at `fd31401` (checked 2026-09-15).

Found by an AI assistant working with a human who was fuzz-testing their own application; reproduced in a minimal project with default settings.

## Request

`listPosts(first: 0)`, and `listPosts(last: 0, before: "<valid cursor>")`. A relationship field's `comments(limit: 0)` returns an empty list normally.

## Observed

`"Something went wrong. Unique error id: ..."` with
```
** (Ash.Error.Unknown)
* ** (Spark.Options.ValidationError) invalid value for :page option: invalid value for :limit option: expected positive integer, got: 0
```
in the log.

## Expected

A rendered error, or an empty page for 0 (the Relay connection spec treats `first: 0` as an empty connection).

## Failure point

`validate_keyset_opts/3`, [lib/graphql/resolver.ex#L1345-L1349](https://github.com/ash-project/ash_graphql/blob/v1.11.0/lib/graphql/resolver.ex#L1345-L1349).

## Notes

Two defects stack here: `first` / `last` are mapped onto the page `limit` without validation, and Ash surfaces the resulting `Spark.Options.ValidationError` as `Ash.Error.Unknown` (reported separately as `ash/invalid-page-options-unrendered`). The ash_graphql half is the one this report is about; the negative page size report shares it.

## Reproduction

* Description, with the failure point at this release: https://github.com/grempe/ash-fuzz-repros#ash_graphqlzero-page-size-unrendered
* Failing test: https://github.com/grempe/ash-fuzz-repros/blob/main/test/ash_graphql/zero_page_size_test.exs
* Run: `mix setup && mix test test/ash_graphql/zero_page_size_test.exs` in https://github.com/grempe/ash-fuzz-repros

PR #476, opened 2026-09-16
fix: reject zero and negative page sizes with a rendered error

### LINKED ISSUE #471: A negative page size makes complexity analysis raise
Versions: ash_graphql 1.11.0, absinthe 1.12.0, absinthe_plug 1.5.10, ash 3.33.3, ash_postgres 2.13.1, Elixir 1.20.1 on OTP 29. Present on the default branch at `fd31401` (checked 2026-09-15).

Found by an AI assistant working with a human who was fuzz-testing their own application; reproduced in a minimal project with default settings.

## Request

With `analyze_complexity: true` on `Absinthe.Plug` (the configuration in ash_graphql's own documentation): `listPosts(first: -1)`, `listPosts(last: -1)`, or a relationship field `comments(limit: -1)`.

## Observed

The request raises before any resolver runs (a 500 with no GraphQL error body):
```
** (Absinthe.AnalysisError) Invalid value returned from complexity analyzer.
Analyzing field: listPosts ... Got value: -2
The complexity value must be a non negative integer.
```
Setting `max_complexity` does not change this; the raise happens in Absinthe's analysis phase before the limit is read. With `analyze_complexity: false`, `listPosts(first: -1)` is the same unrendered `Spark.Options.ValidationError` as `first: 0` (see zero-page-size-unrendered), while `comments(limit: -1)` returns an empty list.

## Expected

A GraphQL error for the invalid page size, and a complexity value that is never negative.

## Failure point

`AshGraphql.Graphql.Resolver.query_complexity/3`, [lib/graphql/resolver.ex#L3498-L3538](https://github.com/ash-project/ash_graphql/blob/v1.11.0/lib/graphql/resolver.ex#L3498-L3538); Absinthe raises for a negative return at [lib/absinthe/phase/document/complexity/analysis.ex#L88-L95](https://github.com/absinthe-graphql/absinthe/blob/v1.12.0/lib/absinthe/phase/document/complexity/analysis.ex#L88-L95).

## Reproduction

* Description, with the failure point at this release: https://github.com/grempe/ash-fuzz-repros#ash_graphqlnegative-page-size-complexity
* Failing test: https://github.com/grempe/ash-fuzz-repros/blob/main/test/ash_graphql/negative_page_size_complexity_test.exs
* Run: `mix setup && mix test test/ash_graphql/negative_page_size_complexity_test.exs` in https://github.com/grempe/ash-fuzz-repros

grempe:
Closes #471 and #474.

Both issues are one gap at the GraphQL boundary. `first`, `last` and `limit` are plain `Int` arguments, so zero and negative values pass schema validation:

- A zero page size reached `Ash.read/2`, whose page option validation failed with a `Spark.Options.ValidationError` wrapped in `Ash.Error.Unknown`, so the client saw "Something went wrong" (#474).
- A negative page size never got that far. `query_complexity/3` multiplied it into a negative complexity and Absinthe raised `Absinthe.AnalysisError` before resolution (#471). Clamping the complexity alone just turns #471 into #474.

Changes in `AshGraphql.Graphql.Resolver`:

- `validate_offset_opts/3` and `validate_keyset_opts/3` return an `Ash.Error.Query.InvalidQuery` for a `limit`, `first` or `last` below one, in the same shape as the existing `first`/`last`/`before`/`after` combination errors, so the client gets a rendered `invalid_query` error.
- `apply_load_arguments/5` adds the same error to the nested query for a negative `limit` on a relationship field. Ash accepts any integer there, so this previously returned an empty list without complaint. A `limit` of zero on a relationship still returns an empty list. This is the one behaviour change that is a judgement call; happy to drop it if you would rather leave relationship arguments to Ash.
- The three `query_complexity/3` clauses clamp the page size at zero, so complexity analysis never returns a negative value.

Tests: new `test/page_size_test.exs` covers zero and negative `limit` (offset), `first` and `last` (keyset), a negative page size under `analyze_complexity: true`, and the relationship `limit` cases. On `main`, the offset and keyset cases return the unrendered Unknown error, the complexity case raises `Absinthe.AnalysisError`, and the relationship case returns an empty list. Full suite 444/444 on `main` and 429/429 with the patch applied to v1.11.0. `mix format --check-formatted`, `mix credo --strict`, `mix sobelow` and `mix dialyzer` clean.

Found by an AI agent working with a human fuzz-testing their own application; the reproductions outside this repo are https://github.com/grempe/ash-fuzz-repros/blob/main/test/ash_graphql/negative_page_size_complexity_test.exs and https://github.com/grempe/ash-fuzz-repros/blob/main/test/ash_graphql/zero_page_size_test.exs.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

